PR #1, opened 2017-05-30
State

jhford:
Here's the first bit of code.  This is a basic State tracking class which has some useful functions.  For the time being, I'm mainly concerned about tracking in-flight information in the database and nothing else.  Metrics should be done in the business logic, outside of this class.

@imbstack I was hoping that you would have a chance to look at this code, particularly around the queries.  I'm no Postgres expert, but I *think* i've written the queries correctly.  Let me know if you think there's any tests we should implement.


sql/create-db.sql
```diff
@@ -0,0 +1,76 @@
+-- Set up a PostgreSQL Database for use as the backing
```
imbstack: These seem correct to me. I've always been wary of triggers because somebody told me to be wary of them 10000 years ago. I don't have much empirical evidence for that now other than that they're kinda spooky-action-at-a-distance if you're just reading the code.

Do we have a plan for migrations and that sort of thing?

lib/main.js
```diff
@@ -0,0 +1,167 @@
...
+
+const types = PG.types;
+
+// Since Postgres returns 64-bit ints from certain operations,
+// we want to parse them down to a JS number.  Yes, we're loosing
+// precision here, but the chances of us having any integer that's
+// greater than ~53-bits here tiny
+types.setTypeParser(20, x => parseInt(x))
+
+//PG.defaults.ssl = true;
```
gregarndt: perhaps we conditionally drop this if the profile is "development" just to ensure it's always enabled in production

lib/state.js
```diff
@@ -0,0 +1,512 @@
...
+    assert(typeof region === 'string');
+    assert(typeof instanceType === 'string');
+    assert(typeof id === 'string');
+    assert(typeof state === 'string');
+    assert(typeof status === 'string');
+
+    let text = [
+      "INSERT INTO spotrequests (id, workerType, region, instanceType, state, status)",
+      "VALUES ($1, $2, $3, $4, $5, $6)",
+      "ON CONFLICT (region, id) DO UPDATE SET state = $5, status = $6;"
```
gregarndt: so anywhere that we're doing an upsert, is there any value in also putting some guards around updating the state/status only if they are certain conditions to begin with? I haven't read through everything just yet, so didn't know if tehre could be updates done out of order

jhford:
@imbstack i'm addressing feedback in a different branch, so this PR might not have anything show up that they're address, but I'm not ignoring the feedback :)

README.md
```diff
@@ -0,0 +1,46 @@
...
+  * `requestSpotInstance(workerType, launchSpecification)` -- Request a spot instance.   This will ensure that the instance and spot request are associated with the workerType and will take
+     the return value of the actual EC2 call to insert into the database as the point of record
+  * `cancelSpotRequest(spotInstanceRequestId)` -- Request that a spot instance request is cancelled
+  * `killInstance(instanceId)` -- Request that an instance is killed
+  * `getState(workerType)` -- Return a view of state for a given worker type
+
+When `requestSpotInstance` is called, the response will be stored in the State Database to store the static values.  We'll have to store the spot request id in a list of ids to continuously poll the describeSpotInstanceRequests api, but because it'll only be until either the request is fulfilled or rejected, we'll be able to poll it way less.
+
+### Setup script
+This script will ensure that there's a properly confirgured rule named `ec2-event-state-transitions` which has a target `ec2-event-queue` which is an SQS Queue in each of the regions.
```
jhford: ty

lib/aws-request.js
```diff
@@ -0,0 +1,96 @@
...
+  let request;
+
+  try {
+    // We have to have a reference to the AWS.Request object
+    // because we'll later need to refer to its .response property
+    // to find out the requestId. part 1/2 of the hack
+    request = service[method](body);
+    let response = await Promise.race([
+      request.promise(),
+      delayer(240 * 1000)().then(() => {
```
jhford: yep!  it is in the new branch

lib/aws-request.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+    if (!err.service) {
+      err.service = serviceId;
+    }
+    if (!err.method) {
+      err.method = method;
+    }
+
+    logObj.msg = 'aws request failure';
+    console.log(JSON.stringify(logObj));
```
jhford: yes! I was thinking maybe next week we could work together on adding monitoring, I'm not entirely sure which datapoints we care about

sql/create-db.sql
jhford: the triggers are basically to implement a TTL system for EC2 Ids, since they are not universally unique and can be reused.  Basically, I want to have something that says anything older than 1 week is absolutely, 100% unexpected.

I don't have strong thoughts on triggers, but am also not a postgres expert.  This could be a bad idea for sure!

sql/create-db.sql
```diff
@@ -0,0 +1,76 @@
...
+  PRIMARY KEY(id, region)
+);
+
+-- instances table contains minimal information on
+-- any instances owned by this ec2 manager.  We don't reference
+-- the spotrequest table because the application logic will be
+-- required to delete any spotrequests which are outstanding
+CREATE TABLE IF NOT EXISTS instances (
+  id VARCHAR(128) NOT NULL, -- opaque ID per Amazon
+  workerType VARCHAR(128) NOT NULL, -- taskcluster worker type
```
jhford: that'd be OK for me!  i'll try to do that change next week

lib/main.js
jhford: so that's actually commented because it didn't work after I switched to using the `pg-pool` module.  I will make sure that SSL is enabled where appropriate, but since this is running on an internal heroku network, SSL is not *superduper critical*, but obviously preferable.

sql/create-db.sql
imbstack: Ok, sounds good. Maybe we should find an expert and quiz them sometime in the next couple weeks.

imbstack:
Sounds good to me! This is looking good.

lib/state.js
jhford: So the only place where I foresee actually using the upsert is in the event listener for the cloud-watch events messages. The upsert is safe providing that the messages are received in correct order.  The problem is that I'm not really sure how certain we could be about this edge case, or how we'd reliably mitigate it.

The underlying issue is that the messages are being received from the CloudWatch Events system, which I think is reasonable to assume is massively distributed.  My assumption is that Cloud-watch events internal putEvent calls are done in the correct order, and thus end up in the sqs queue in correct order.  Because of the system being massively distributed, doing something like comparing the timestamps of the messages would be very complicated and likely not result in a much stronger guarantee.

gregarndt: ok, I was thinking in the case where you might have multiple consumers listening to the same event stream but if that's not the case here, I don't see a problem.

gregarndt:
@jhford Any chance you could push that branch up to the repo so we can check it out?

PR #2, opened 2017-06-09
Cloud watch listening

jhford:
This addresses most of the feedback from the first PR.  There's a couple issues with the spot request polling that I need to finish writing tests for, and also shake out with spot requests being sent in from the provisioner.  That's not currently happening because the patch to send the spot requests to the ec2-manager was backed out.


jhford:
I figured I'd push the latest code to the branch.  Please note the following things:

1. unit tests are missing for requestSpotInstance -- i am not done them yet
2. the cw listener needs to start checking that the keyPrefix is the first substring of the KeyName value of all describeInstances calls

lib/api.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+async function handleRequestSpotInstanceResponse({state, region, response}) {
+  assert(typeof region === 'string');
+  assert(typeof response === 'object');
+  assert(Array.isArray(response.SpotInstanceRequests));
+  assert(response.SpotInstanceRequests.length === 1);
+  let [spotRequest] = response.SpotInstanceRequests;
+
+  let id = spotRequest.SpotInstanceRequestId;
+  let workerType = spotRequest.LaunchSpecification.KeyName.split(':').slice(1, 2)[0];
```
jhford: because.... umm..... that's a good point :)

PR #3, opened 2017-06-15
More updates to the API

jhford:


jhford:
Hey @imbstack I've asked for review on the provisioner, this is the code that the provisioner changes run against.  I'm not totally done with this branch, but if you have some time to take a look at it I would appreciate it.

lib/zombie-killer.js
```diff
@@ -0,0 +1,76 @@
+const assert = require('assert');
```
imbstack: This makes sense to me!

lib/api.js
```diff
@@ -466,6 +500,16 @@ api.declare({
 
 });
 
+/*****************************************************************************/
+/*****************************************************************************/
+/*    NOTE:  ALL FOLLOWING METHODS ARE INTERNAL ONLY AND ARE NOT             */
+/*           INTENDED FOR GENERAL USAGE.  AS SUCH THEY ARE ALL               */
+/*           CONSIDERED TO BE EXPERIMENTAL, DO NOT HAVE ANY                  */
+/*           SCHEMA DESCRIPTIONS AND ARE INTENDED TO BE CHANGED              */
```
imbstack: This makes me a little sad. I get that it's not a problem in this case, but it's a weird precedent to set sorta?

PR #13, opened 2017-06-26
insert cloud watch events into the database

jhford:
This is mainly for debugging right now, but it is an interesting data set which we might be able to use later.  Currently the table only has instance id, region, state and the time the event was received, but could be expanded upon later


jhford:
The main commit here is https://github.com/taskcluster/ec2-manager/pull/13/commits/04ee5aebde06728050edfa5690e94a156837743c

Brian, please merge the PR if it looks OK to you.

PR #16, opened 2017-07-03
Store more information in the state tables (resolves #4)

### LINKED ISSUE #4: Add more fields to the database for UI
The following fields should be added to the Database schema.  This will require updates to any code which inserts data into the database as well as a lot of tests

Instances [(Reference)](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeInstances-property):
- `instance.ImageId` -- `varchar(128)` column should be called `imageid`
- `instance.LaunchTime` -- `TIMESTAMPTZ` column should be called `launched`
- `instance.Placement.AvailabilityZone` -- `varchar(128)` column should be called `az`

Spot Requests[(Reference)](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeSpotInstanceRequests-property):
- `spotRequest.LaunchSpecification.ImageId` -- `varchar(128)` column should be called `imageid`
- `spotRequest.CreateTime` -- `TIMESTAMPTZ` column should be called `created`
- `spotRequest.LaunchSpecification.Placement.AvailabilityZone` -- `varchar(128)` column should be called `az`

These fields should be used to insert real values in the `AwsManager.stateForStorage()` method instead of the current `'add-me'` value.

jhford:
This PR adds AvailabilityZone, ImageId and Launch or Create time for state entries.  These are the fields which we'd like to show in the provisioner UI.

### DIFF
--- a/lib/cloud-watch-event-listener.js
+++ b/lib/cloud-watch-event-listener.js
@@ -143,6 +143,9 @@ class CloudWatchEventListener extends events.EventEmitter {
       let [provisionerId, workerType] = instance.KeyName.split(':');
       let instanceType = instance.InstanceType;
       let srid = instance.SpotInstanceRequestId;
+      let az = instance.Placement.AvailabilityZone;
+      let imageId = instance.ImageId;
+      let launched = new Date(instance.LaunchTime);
 
       // We check for workertype being truthy because it's always possible that
       // this instance is one which is not in the provisioner/ec2-manager sphere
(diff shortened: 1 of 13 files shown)


jhford:
```
ec2-manager-staging::DATABASE=> select * from instances order by launched;
         id          | workertype |  region   | instancetype |  state  |     srid     |            touched            |     az     |   imageid    |        launched        
---------------------+------------+-----------+--------------+---------+--------------+-------------------------------+------------+--------------+------------------------
 i-07724a70e3b7d4c55 | cli        | us-east-1 | r3.xlarge    | running | sir-sqvr6b3j | 2017-07-03 14:10:54.284802+00 | us-east-1c | ami-2f15f642 | 2017-07-03 13:55:42+00
```

Example row from staging.  I used the `ALTER` sql script to upgrade the staging database.

lib/cloud-watch-event-listener.js
```diff
@@ -156,31 +159,27 @@ class CloudWatchEventListener extends events.EventEmitter {
...
       }
+      let opts = {
+        workerType,
+        region,
+        az,
+        instanceType,
+        imageId,
+        id,
+        state,
+        srid: srid ? srid : 'undefined',
```
imbstack: I prefer `srid || 'undefined'` in this sort of case generally. Also maybe using `'undefined'` the string is a bit confusing since it is a concept in js? Both are just nits.

lib/state.js
```diff
@@ -163,21 +255,26 @@ class State {
...
+      id,
+      workerType,
+      region,
+      az,
+      instanceType,
+      state,
+      status,
+      imageId,
+      created,
+    } = assertValidSpotRequest(spotRequest);
```
imbstack: Nice cleanup.

lib/cloud-watch-event-listener.js
jhford: i'll change this after i merge #17 so i don't need to rebase that patchset

jhford: and for the time being, there's nothing stored which does not have an srid.

lib/state.js
jhford: thanks!

Issue #9, opened 2017-06-22
Consider fetching immutable fields in the CloudWatchEventListener before using the API


jhford:
There are certain fields in the database which contain information which is immutable.  For example, the worker type that an instance counts for should never change.  In the case of any EC2 State Change event, we should try to load the data for that `{region, id}` row and use it instead of needing to run a `describeInstances` request.

This is not too high priority but would theoretically cut the number of `describeInstances` requests we make in half.  This would also let us log more information for the cases which result in instance deletion, but should not become mandatory.  The reason I want to avoid this requirement is that processing deletions is nearly free, ec2-api-wise.

PR #17, opened 2017-07-04
Ensure ordered delivery

jhford:
This branch ensures that once a resource is tracked in the state database, all future events only mutate state if they were sent at a time *later* than the state stored in the database.  For those messages which are older than what's stored in the database, we simply ignore those messages.


jhford:
This resolves [bug 1376034](https://bugzilla.mozilla.org/show_bug.cgi?id=1376034) as well as #9.

When this branch is merged, the provisioner UI will also show values where it currently shows `add-me`.

lib/cloud-watch-event-listener.js
```diff
@@ -104,15 +104,104 @@ class CloudWatchEventListener extends events.EventEmitter {
...
+        imageId: dbResponse.imageId,
+        launched: dbResponse.launched,
+        state,
+        lastevent: generated,
+        metadataSource: 'db',
+      };
+      
+      if (dbResponse.lastevent < generated) {
+        try {
+          if (state === 'pending' || state === 'running') {
```
gregarndt: Do we have a race here where an instance could be removed from the table if it's in a "rebooting" state, but then we try to spin up more instances because it looks like we have less running capacity than what we might really have? Some windows worker types reboot after every task, and at least one during the initial setup before claiming work.

gregarndt: maybe this is a good spot to log stats of instances entering a $state

lib/cloud-watch-event-listener.js
```diff
@@ -169,15 +258,25 @@ class CloudWatchEventListener extends events.EventEmitter {
...
         log.debug(opts, 'Ignoring instance because it does not belong to this manager');
       }
     } else {
+      // For those events which are for instances which aren't being tracked in
+      // the database, we'll just blindly delete them.  This might have
+      // overhead on the database, but it does ensure that they're definitely
+      // removed from state.  The cost is likely a non-issue since the only way
+      // to avoid the removal would be the much more costly EC2 api calls to
+      // figure out metadata.  If we wanted to start tracking instance outcome,
+      // we could do it here (and also in the dbResponse truthy clause above).
```
gregarndt: ah, maybe this comment here addresses my comment earlier about logging stats.  If we're putting guards around state and know what state an instance is, perhaps this is the best time to add those stats rather than waiting later.

lib/state.js
```diff
@@ -220,6 +230,41 @@ class State {
...
 
   /**
+   * Begin a transaction
+   */
+  async beginTransaction() {
+    let client = await this.getClient();
+    try {
+      await client.query("BEGIN;");
+      return client;
+    } catch (err) {
```
gregarndt: I think if an error is raised here and no client is returned, then there could be some unintended side effects, or at least the caller thinks it' operation on a transaction when it's really not.  beginTransaction will return undefined to the caller, and then that is passed onto other methods thinking it has a transaction.  In some cases, if undefined is passed in for the client, such as with _listTable, then it will just do a query without a transaction I think.  If we are explicitly wanting a transaction for an operation, we probably should verify that a client was returned and there was not an error thrown.

lib/cloud-watch-event-listener.js
```diff
@@ -104,15 +104,104 @@ class CloudWatchEventListener extends events.EventEmitter {
...
+    // Not a great name, but this variable is the time that the CloudWatch
+    // Event message was generated by the API, not the time at which we
+    // received it
+    let generated = new Date(body.time);
 
+    // We're logging this information to a database table so that we can
+    // see when these events happen
     try {
-      await this.state.logCloudWatchEvent({region, id, state});
+      await this.state.logCloudWatchEvent({region, id, state, received: generated});
```
jhford: I'd agree.
`commit 85ba0bc2ead91fb70d4358a77e35719fd6e9f840`

lib/cloud-watch-event-listener.js
```diff
@@ -104,15 +104,104 @@ class CloudWatchEventListener extends events.EventEmitter {
     let region = body.region;
     let id = body.detail['instance-id'];
     let state = body.detail.state;
+    // Not a great name, but this variable is the time that the CloudWatch
+    // Event message was generated by the API, not the time at which we
+    // received it
+    let generated = new Date(body.time);
 
+    // We're logging this information to a database table so that we can
```
jhford: sure, part of `commit 85ba0bc2ead91fb70d4358a77e35719fd6e9f840`

lib/cloud-watch-event-listener.js
jhford: So before we go too far into the discussion, I think it's worth noting that we don't get sent rebooting events.  Whether this is intentional, I'm not 100% sure.

```
ec2-manager-staging::DATABASE=> select distinct state from cloudwatchlog;
     state     
---------------
 pending
 running
 stopping
 shutting-down
 stopped
 terminated
(6 rows)
```
This is all instances in the ec2 account, not just staging ones.

I don't know if I would consider this to be a race condition, the instance *is* out of the pool during the time that it's rebooting, and even if `rebooting` events were sent, there would logically be a corresponding `running` event which would put it back into the calculations as running.

jhford: yep, we're doing that with the `cloudwatchlog` table.  I've added a `received` column to the cloudwatchlog table which will keep track of drift between api time and msg generation time

lib/cloud-watch-event-listener.js
jhford: this is likely where that code will live

lib/state.js
jhford: yep, i should be rethrowing errors in these functions.

jhford: `[ensure-ordered-delivery 74193a2] Rethrow errors when transactions are created/commited/rolledback
`

lib/cloud-watch-event-listener.js
```diff
@@ -104,15 +104,102 @@ class CloudWatchEventListener extends events.EventEmitter {
...
     }
 
-    if (state === 'pending' || state === 'running') {
+    let dbResponse;
+
+    let transaction = await this.state.beginTransaction();
+
+    try {
+      let x = await this.state.listInstances({region, id}, transaction);
+      assert(Array.isArray(x));
```
imbstack: Is this necessary?

jhford: not really, but I prefer to over-assert so that errors are detected as early a call chain as possible

Issue #8, opened 2017-06-22
Move pricing into the ec2-manager


jhford:
The AWS Provisioner still maintains the spot pricing information system.  Let's move this knowledge into the EC2-Manager.  There should be a background poller which checks the pricing data every X minutes (configurable) along with an endpoint which can return information on pricing.  The endpoint should be `/v1/pricing` and take a request body in the format
```javascript
[
  {region, instanceType, count},
]
```
and return a list of prices in the format:

```javascript
[ {region, instanceType, az, price, min, max, preference} ]
```

The preference values should be a number between 1 and 1.000.000 which shows ec2-manager internal preference.  1 should be the least preferred option.  This value will be used to show things that the EC2 Manager wishes to convey, for instance there might be too many instances in a region and the manager would prefer more in other regions even though it will allow that configuration.  Min and Max should be the minimum and maximum number of instances that this information is considered to be valid for

PR #18, opened 2017-07-10
Pricing in EC2-Manager

jhford:
I'd like to start being able to move pricing out of the aws-provisioner codebase and into the EC2 Manager codebase.

This PR accomplishes that.


jhford:
This would resolve #8

lib/api.js
```diff
@@ -512,6 +513,45 @@ api.declare({
...
+  title: 'Request prices for EC2',
+  stability: API.stability.experimental,
+  input: 'prices-request.json#',
+  output: 'prices.json#',
+  description: [
+    'Return a list of possible prices for EC2'
+  ].join(' '),
+}, async function (req, res) {
+  let prices = await this.pricing.getPrices(req.body);
+  console.dir(req.body);
```
imbstack: Leftover debug?

lib/main.js
```diff
@@ -245,6 +261,7 @@ let load = loader({
             try {
               await pollSpotRequests({ec2: ec2[region], region, state});
             } catch (err) {
+              console.dir(err.stack || err);
```
imbstack: Debug?

lib/main.js
```diff
@@ -299,13 +316,40 @@ let load = loader({
...
+        credentials: cfg.taskcluster.credentials,
+      });
+
+    },
+  },
+
+  purgequeues: {
+    requires: ['server', 'apiClient'],
+    setup: async ({server, apiClient}) => {
+      await apiClient.purgeQueues();
```
imbstack: Am I reading this correctly that we're calling a method on ourselves via http? Why not do this work directly?

lib/pricing-poller.js
```diff
@@ -0,0 +1,242 @@
...
+    return this.iterator.stop();
+  }
+
+  /**
+   * Take a list of {price, time} and return a price that we should use
+   */
+  _findSpotPriceForInstanceType(series) {
+    let now = new Date();
+    let maxPrice = 0;
+    for (let {price, time} of series) {
```
imbstack: Are you in the anti lodash camp? If not, we can replace this loop with `_.maxBy(series, 'price')` maybe?

jhford:
So calling ourselves by http is more to implement a simple script to purge the queues, not something we're doing in production.

lib/pricing-poller.js
jhford: I'd rather not use lodash

lib/main.js
jhford: this is just a shortcut for development

lib/main.js
jhford: I'm not using the debug module, but I guess using `taskcluster-lib-log` would be better here.

lib/api.js
jhford: yep!

jhford:
I'm going to land this.  The purgequeues stuff is purely a convenience for development, i'd rather remove it altogether than figure out a better way to do it.

jhford:
rebased with:
```
commit 9b747a355d8596eb54d72a9220737cd33600bfd9
Author: John Ford <john@johnford.org>
Date:   Mon Jul 10 19:26:39 2017 +0200

    Add pricing to the EC2-Manager API

commit 7789ad97cd89df6b27472bed6ba6f6c137d6ad3b
Author: John Ford <john@johnford.org>
Date:   Mon Jul 10 13:54:34 2017 +0200

    Code which fetches pricing.  Currently limited to spot-pricing

```

PR #21, opened 2017-07-20
Create indices for workertypes on both tables

jhford:
We search by worker type fairly commonly, so I would like to index that value


jhford:
the new `sorter` function is required because using the index changes the ordering of the entries inside of the database.  The arrays aren't ordered, but we don't really have a way to represent a set easily here.

petemoore:
Sorry I didn't do a good job of reviewing here. 😞

Issue #12, opened 2017-06-22
Implement the runInstances EC2 API


jhford:
We should support running instances using the `runInstances` method as well as by requesting a spot request.  This would enable us to use On-Demand instances and would be nice because all the metadata we need for the internal state tables would already be in the response value from the EC2 api call, which would mostly eliminate the need for the many `describeInstances` calls which we make.

PR #23, opened 2017-07-26
Added skeleton implementation for runInstances API endpoint

petemoore:
Started work on #12.


jhford:
Unfortunately, this patch had bitrotted and I ended up doing it from scratch.  Thanks for the first go!

PR #26, opened 2017-09-22
Refactor spot request polling into class

auni53:
* support multiple regions at a time
* rename spot-request-poll.js to spot-request-poller.js


lib/spot-request-poller.js
```diff
@@ -0,0 +1,116 @@
...
+const {runAWSRequest} = require('./aws-request');
+const Iterate = require('taskcluster-lib-iterate');
+
+const log = require('./log');
+
+const BATCH_SIZE = 100;
+
+class SpotRequestPoller {
+
+  constructor({ec2, regions, runaws = runAWSRequest, spotPollDelay = 25, state}) {
```
rstiyer: Where does the default value for spotPollDelay come from? Why 25, in particular?

lib/spot-request-poller.js
```diff
@@ -0,0 +1,116 @@
...
+          log.info(infoObj, 'No longer tracking spot request');
+        }
+      }
+
+      if (idsToKill.length > 0) {
+        let killed = await this.runaws(this.ec2, 'cancelSpotInstanceRequests', {
+          SpotInstanceRequestIds: idsToKill,
+        });
+        let idsKilled = killed.CancelledSpotInstanceRequests.map(x => {
+          return x.SpotInstanceRequestId
```
rafvasq: Could use arrow function, i.e. `x => x.SpotInstanceRequestId`

lib/spot-request-poller.js
```diff
@@ -0,0 +1,116 @@
...
+        });
+        await Promise.all(idsKilled.map(id => {
+          return this.state.removeSpotRequest({region, id})
```
rafvasq: Could use arrow function, i.e. `id => this.state.removeSpotRequest({region, id})`

lib/spot-request-poller.js
auni53: delivered to me in a dream by divine spirits

auni53: jk https://github.com/taskcluster/ec2-manager/blob/master/config.yml#L5

rstiyer: Do you need to specify a default value if it's already been set in config.yml? I think that would already be taken into account by passing cfg.app.spotPollDelay

rstiyer: 🖖

lib/spot-request-poller.js
```diff
@@ -0,0 +1,113 @@
...
+  stop() {
+    return this.iterator.stop();
+  }
+
+  async poll() {
+    await Promise.all(this.regions.map(region => this._poll(region)));
+  }
+
+  async _poll(region) {
+    let idsToPoll = await this.state.spotRequestsToPoll({region});
```
jhford: can we do an assertion that region is a string?

lib/spot-request-poller.js
```diff
@@ -0,0 +1,113 @@
...
+const log = require('./log');
+
+const BATCH_SIZE = 100;
+
+class SpotRequestPoller {
+
+  constructor({ec2, regions, runaws = runAWSRequest, spotPollDelay = 25, state}) {
+    assert(typeof ec2 === 'object');
+    assert(Array.isArray(regions));
+    regions.forEach(region => assert(typeof region == 'string'));
```
auni53: o this should be ===

lib/spot-request-poller.js
auni53: We do that [here](https://github.com/taskcluster/ec2-manager/pull/26/files/3ed978da12351bf42682993634fea1b4c641b4d2#diff-bb870db80b11d148f2a4f0445f4bbfaaR14); should we double-check it?

PR #28, opened 2017-09-28
allow manage-worker-type permissions to run terminateInstance()

jhford:
So the idea here is to allow someone to kill an instance based on a per-workertype scope instead of allowing an all-or-nothing approach.  The downside here is that we have to do a DB query to determine whether the request is allowed or not, but I think this should be OK.

My concern is mainly about the behaviour of passing in an undefined value into the req.satisfies method.

This should help @helfi92 implement the kill instance action.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -409,15 +409,26 @@ api.declare({
   name: 'terminateInstance',
   title: 'Terminate an instance',
   stability: API.stability.experimental,
-  scopes: [['ec2-manager:manage-instances:<region>:<instanceId>']],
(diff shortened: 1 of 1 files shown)


helfi92:
@jhford Does this need to be merged? The action button is starting to be used by people like sherrifs now.

jhford:
It will need to be merged and deployed.  I'm hoping to do a deployment early next week.  In the meantime, we can grant the `ec2-manager:manage-instances:*` scope to those for which this is important and relevant.

PR #29, opened 2017-09-30
Add table for ami usage

auni53:
(paired w @rafvasq so you can ask him questions too)

1. Is the `amiusage` table name fine?

2. Do we need any tests to assert the table is created? (Editing `lib/state.js` will come in a future commit.)

3. @jhford I can't add reviewers

### DIFF
--- a/sql/create-db.sql
+++ b/sql/create-db.sql
@@ -85,3 +85,10 @@ CREATE TABLE IF NOT EXISTS cloudwatchlog (
   received TIMESTAMPTZ NOT NULL DEFAULT NOW(),
   PRIMARY KEY (id, region, state, generated)
 );
+
+-- Amazon Machine Image (ami) usage
+CREATE TABLE IF NOT EXISTS amiusage (
+  region VARCHAR(128) NOT NULL, -- ec2 region
+  id VARCHAR(128) NOT NULL,
+  lastused TIMESTAMPTZ NOT NULL,
+);
--- a/sql/clear-db.sql
+++ b/sql/clear-db.sql
@@ -2,3 +2,4 @@
 DELETE FROM spotrequests;
 DELETE FROM instances;
 DELETE FROM cloudwatchlog;
+DELETE FROM amiusage;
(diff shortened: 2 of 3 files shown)


sql/create-db.sql
```diff
@@ -85,3 +85,11 @@ CREATE TABLE IF NOT EXISTS cloudwatchlog (
   received TIMESTAMPTZ NOT NULL DEFAULT NOW(),
   PRIMARY KEY (id, region, state, generated)
 );
+
+-- Cloudwatch Events Log
```
jhford: We should update the comments here too

sql/create-db.sql
```diff
@@ -85,3 +85,11 @@ CREATE TABLE IF NOT EXISTS cloudwatchlog (
...
+-- We want to keep a log of when every cloud watch event was generated
+CREATE TABLE IF NOT EXISTS amiusage (
+  region VARCHAR(128) NOT NULL, -- ec2 region
```
walac: Not sure if there is any influence in database performance, but 128 chars for regions is way more than we need.

auni53: Ah, was following the spec @jhford for this

jhford: Yep, we're using `VARCHAR(128)` everywhere for identifiers from EC2 so that we can just use their identifiers verbatim.  We use something fairly large because we're not actually using more space in the DB and it lets us handle changes in identifier sizes without any schema changes.  This happened this or last year with the identifiers nearly doubling in size.

from https://www.postgresql.org/docs/9.6/static/datatype-character.html refering the difference between character types:

> Tip: There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

auni53:
@jhford could you give another review and merge this in?

PR #30, opened 2017-10-05
[30% Review] Request info about EBS volumes

SamanthaYu:
@rstiyer and I implemented a function that fetches the EBS volume data. However, we are not sure how to implement the unit tests.

@jhford @imbstack @djmitche


rstiyer:
The problem, as we see it, is that the function we are implementing (which fetches the volume data from EC2) probably shouldn't modify the return of the sweep function in housekeeping.js, since that function currently only returns information about "problematic" instances. However, this makes it difficult for us to test the implementation of the new function. 

@imbstack suggested that we add a parameter to the housekeeping constructor to specify the secondary function we will call on the paginated response, so we could easily mock it in the tests.

Is this an appropriate approach? It would certainly complicate the housekeeping class, but it seems like the most straight-forward way to allow direct unit testing of the new EBS monitoring function.

lib/housekeeping.js
```diff
@@ -67,6 +82,8 @@ class HouseKeeper {
       let stateInstanceIds = stateInstances.map(x => x.id);
       let stateRequestIds = stateRequests.map(x => x.id);
 
+      sweepVolumes();
```
jhford: Since this is a promise, we should either store the promise and await it later on, or we should await it now.

lib/housekeeping.js
```diff
@@ -38,9 +38,24 @@ class HouseKeeper {
...
+  async sweepVolumes() {
+    let volumes = await this.runaws(this.ec2[region], 'describeVolumes', {
+      Filters: [{
+        Name: 'status',
+        Values: ['available', 'in-use'],
+      }],
+    });
+
+    for (let volume of volumes.Volumes) {
+      handleVolumeData(volume);
```
djmitche: handleVolumeData is async, and needs to be awaited.

djmitche: Also, if we're doing anything potentially time-consuming in `handleVolumeData`, then we should probably call it in parallel for each returned volume:
```
await Promise.all(volumes.Volumes.map(handleVolumeData));
```
This calls `handleVolumeData` for each element of the array, then collects all of the resulting Promises and waits until they are all complete (or one fails).

rstiyer:
Thanks, @jhford and @djmitche, for the reviews! We missed the async business in the first run through, so thanks for pointing that out.

I understand what you mean about testing the function sweepVolumes() on its own; I guess we were hesitant to do that because it seemed like we should be treating it as a private method to housekeeping.js. Thanks for pointing us in the right direction!

lib/housekeeping.js
walac: I would vote for the `Promise.all` approach anyway. Easier to read, easier to write and more effient.

SamanthaYu:
Thanks everyone for all the help and clarification!
I'm closing this PR to clean it up. Either @rstiyer or I will create a new PR with the above changes.

walac:
> I'm closing this PR to clean it up. Either @rstiyer or I will create a new PR with the above changes.

You don't need to close the PR, just pushing new commits will automatically update the PR. This is even recommended as it preserves the discussion around the patches.

djmitche:
I recommended closing, as this was a 30% review and squashed in a bunch of commits from @auni53's team.  A clean slate will simplify the git fixing.

PR #31, opened 2017-10-11
Create sweepVolumes method to fetch and process EBS volumes data

rstiyer:
We implemented the recommended changes from Samantha's earlier pull request:
* Call sweepVolumes with the await operator
* Encapsulate calls to handleVolumeData within Promise.all(...)

In addition, we added 3 unit tests for the new method, sweepVolumes:
1. Mock sweepVolumes to ensure that it is actually called by sweep()
2. Test that sweepVolumes can handle receiving an empty dataset from AWS. Since the previous test verifies that sweepVolumes is being called by sweep(), we simply run sweep() with an appropriate return value from the describeVolumesStub.
3. Test that sweepVolumes calls handleVolumeData only once per volume. This requires specifying different return values for the describeVolumesStub per region, since the AWS describeVolumes function should only return volumes which belong to a single region.

We'd like your input on the last test, in particular. It's not very pretty, because in order to specify separate return values per region, it's necessary to specify the exact input to describeVolumesStub. To reduce duplication of the parameters to the 'describeVolumes' AWS function-- and because these parameters are unchanging--we created a new field in HouseKeeper to store these parameters. I'm not sure if it's ideal to store these as a global variable, but it definitely makes the testing easier.

One more thought related to the last test: would it make sense to remove the "method" parameter from the describe[...]Stub functions, since we already have an if-else block that checks what method is being requested and returns the appropriate stub? It seems a bit redundant, and the presence of that parameter means that we need to include the parameter when specifying the desired output of the stub.

@djmitche @imbstack @jhford: @SamanthaYu  and I would really appreciate any thoughts or feedback you have. Thanks for all the help so far!

### DIFF
--- a/lib/housekeeping.js
+++ b/lib/housekeeping.js
@@ -40,7 +40,7 @@ class HouseKeeper {
 
   async handleVolumeData(volume) {}
 
-  async sweepVolumes() {
+  async sweepVolumes(region) {
     let volumes = await this.runaws(this.ec2[region], 'describeVolumes', {
       Filters: [{
         Name: 'status',
@@ -48,9 +48,7 @@ class HouseKeeper {
       }],
     });
 
-    for (let volume of volumes.Volumes) {
-      handleVolumeData(volume);
-    }
+    await Promise.all(volumes.Volumes.map(this.handleVolumeData));
   }
 
   async sweep() {
@@ -82,7 +80,7 @@ class HouseKeeper {
       let stateInstanceIds = stateInstances.map(x => x.id);
       let stateRequestIds = stateRequests.map(x => x.id);
 
-      sweepVolumes();
+      await this.sweepVolumes(region);
 
       let instances = await this.runaws(this.ec2[region], 'describeInstances', {
         Filters: [{
--- a/lib/housekeeping.js
+++ b/lib/housekeeping.js
@@ -36,24 +36,19 @@ class HouseKeeper {
     this.maxInstanceLifeHours = maxInstanceLifeHours;
     this.monitor = monitor;
     this.tagger = tagger;
+  }
 
-    this.describeVolumesParams =  {
+  async _handleVolumeData(volume) {}
+
+  async _sweepVolumes(region) {
+    let volumes = await this.runaws(this.ec2[region], 'describeVolumes', {
       Filters: [{
         Name: 'status',
         Values: ['available', 'in-use'],
       }],
-    } 
-  }
+    });
 
-  async handleVolumeData(volume) {}
-
-  async sweepVolumes(region) {
-    let volumes = await this.runaws(
-      this.ec2[region],
-      'describeVolumes',
-      this.describeVolumesParams);
-
-    await Promise.all(volumes.Volumes.map(this.handleVolumeData));
+    await Promise.all(volumes.Volumes.map(this._handleVolumeData));
   }
 
   async sweep() {
@@ -85,7 +80,7 @@ class HouseKeeper {
       let stateInstanceIds = stateInstances.map(x => x.id);
       let stateRequestIds = stateRequests.map(x => x.id);
 
-      await this.sweepVolumes(region);
+      await this._sweepVolumes(region);
 
       let instances = await this.runaws(this.ec2[region], 'describeInstances', {
         Filters: [{
--- a/lib/housekeeping.js
+++ b/lib/housekeeping.js
@@ -38,9 +38,24 @@ class HouseKeeper {
     this.tagger = tagger;
   }
 
+  async handleVolumeData(volume) {}
+
+  async sweepVolumes() {
+    let volumes = await this.runaws(this.ec2[region], 'describeVolumes', {
+      Filters: [{
+        Name: 'status',
+        Values: ['available', 'in-use'],
+      }],
+    });
+
+    for (let volume of volumes.Volumes) {
+      handleVolumeData(volume);
+    }
+  }
+
   async sweep() {
     log.info('starting housekeeping');
-    
+
     /* A Zombie is considered to be an instance which has lived for too long.
      * These instances are often long living because they've forgotten to shut
      * themselves down.  When that happens, we could end up spending a lot of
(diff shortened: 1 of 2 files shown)


lib/housekeeping.js
```diff
@@ -36,11 +36,29 @@ class HouseKeeper {
...
+
+    this.describeVolumesParams =  {
+      Filters: [{
+        Name: 'status',
+        Values: ['available', 'in-use'],
+      }],
+    } 
+  }
+
+  async handleVolumeData(volume) {}
```
jhford: Javascript doesn't really have a built in concept about public/private for object properties.  Because of this, there's a weak convention of using an underscore as the first character of a purely internal property as a hint to the user that such properties shouldn't be used.  I think we should consider using that convention here for those functions which we know are internal to this class.

test/housekeeping_test.js
```diff
@@ -323,5 +350,92 @@ describe('House Keeper', () => {
...
+      Reservations: [
+      ]
+    });
+    
+    describeSpotInstanceRequestsStub.returns({
+      SpotInstanceRequests: [
+      ]
+    });
+
+    describeVolumesStub.withArgs(ec2[region], 'describeVolumes', houseKeeper.describeVolumesParams)
```
jhford: It seems like exposing the `describeVolumesParams` is only needed in order to use here for matching the call.  Maybe instead, we might be able to use the Sinon [matcher api](http://sinonjs.org/releases/v4.0.1/matchers/)?

test/housekeeping_test.js
```diff
@@ -323,5 +350,92 @@ describe('House Keeper', () => {
...
+          CreateTime: new Date().toString(), 
+          Size: 8, 
+          SnapshotId: "snap-1234567890abcdef0", 
+          State: "in-use", 
+          VolumeId: "vol-049df61146c4d7901", 
+          VolumeType: "standard",
+        }]
+      });
+
+    describeVolumesStub.withArgs(ec2['us-east-2'], 'describeVolumes', houseKeeper.describeVolumesParams)
```
jhford: I'm curious why you chose to use `ec2[region]` for the first call, but a specific region for the second one?  I wonder if we should either pick two hard coded regions or do something like `Object.keys(ec2)[0]` and `Object.keys(ec2)[1]`?

lib/housekeeping.js
rstiyer: That seems like a very good idea, thanks John!

test/housekeeping_test.js
```diff
@@ -323,5 +350,92 @@ describe('House Keeper', () => {
       });
     }
   });
-});
 
+  it('should call sweepVolumes exactly once', async() => {
```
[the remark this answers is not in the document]
rstiyer: Yes, it should. Good catch! :+1:

test/housekeeping_test.js
rstiyer: I came up with a solution that is a bit better, I think, but still not perfect. Sinon supports custom matchers, which return true or false to indicate whether the input matches or not. I tested using a custom matcher, and it seems that only the first parameter gets passed along. Luckily we only need to check the value of the first parameter, so it works, but it seems a bit odd.

test/housekeeping_test.js
rstiyer: That's a great point. We wanted to use the global `region` since it was already defined, but wanted to include a second region in the test without creating another global variable. In this case, it would probably make more sense to use two hard-coded keys, since we need both the region string and the corresponding ec2 instance.

test/housekeeping_test.js
```diff
@@ -402,33 +402,35 @@ describe('House Keeper', () => {
...
-      });
+    describeVolumesStub.withArgs(sinon.match(function(value) {
+      return value === ec2['us-west-2'] 
+    })).returns({
+       Volumes: [{
+         Attachments: [],
+         AvailabilityZone: 'us-west-2', 
+         CreateTime: new Date().toString(), 
+         Size: 8, 
+         SnapshotId: "snap-1234567890abcdef0",
```
SamanthaYu: This is probably a minor quibble... SnapshotID and VolumeID are probably unique IDs so we shouldn't be using the same values for both volumes in this test.

We can probably even extract out the returned Volumes object in each call to us-westVolumes and us-eastVolumes. In our future PRs, we could then use those Volume objects in our future tests.

test/housekeeping_test.js
```diff
@@ -402,33 +402,35 @@ describe('House Keeper', () => {
...
-          AvailabilityZone: region, 
-          CreateTime: new Date().toString(), 
-          Size: 8, 
-          SnapshotId: "snap-1234567890abcdef0", 
-          State: "in-use", 
-          VolumeId: "vol-049df61146c4d7901", 
-          VolumeType: "standard",
-        }]
-      });
+    describeVolumesStub.withArgs(sinon.match(function(value) {
```
SamanthaYu: Nice cleanup! 😃

test/housekeeping_test.js
rstiyer: I changed only the last character in both, out of sheer laziness. We should probably make them more unique. :stuck_out_tongue_closed_eyes:

SamanthaYu: And I was too lazy to notice that they were actually different strings. 😆

test/housekeeping_test.js
```diff
@@ -120,6 +127,11 @@ describe('House Keeper', () => {
       ]
     });
 
+    describeVolumesStub.returns({
```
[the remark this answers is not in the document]
rstiyer: I went with your second suggestion for now (setting a default return value), since that's a very quick solution and make intuitive sense to me.

test/housekeeping_test.js
jhford: I have to look at the code to fully understand, but it might be that there's a 1:1 relationship between arguments and matchers?  I'll look at the code

test/housekeeping_test.js
jhford: lazy<sup>lazy</sup>

jhford: :)

jhford:
Can you rebase this onto master again and I'll merge it asap!

Thanks

rstiyer:
@jhford: I didn't get a notification about your previous comment, my apologies. I didn't realize this hadn't been merged yet. I rebased onto master and had to make a few changes to pass linting.

djmitche:
That will auto-deploy to staging.  I'll look at the logging there and if everything seems OK, I'll merge to master.

PR #32, opened 2017-10-14
add reportAmiUsage and ListAmiUsage methods

rafvasq:
Methods added to class State in lib/state.js to interface with table along with unit tests for each. From what I understand, as of right now reportAmiUsage is simply upserting

### DIFF
--- a/lib/state.js
+++ b/lib/state.js
@@ -569,6 +569,30 @@ class State {
 
     return result.rowCount;
   }
+  
+  /**
+   * Either insert or update an AMI's usage.  This ensures an atomic insert or
+   * update, but we never learn which one happened.
+   */
+  async reportAmiUsage({region, id}, client) {
+    assert(typeof region === 'string');
+    assert(typeof id === 'string');
+    let text = [
+      "INSERT INTO amiusage (region, id, lastused)",
+      "VALUES ($1, $2, now())",
+      "ON CONFLICT (region, id) DO UPDATE SET lastused = EXCLUDED.lastused"
+    ].join(' ');
+    
+    let values = [region, id];
+    
+    let result;
+    if (client) {
+      result = await client.query({text, values, name: 'upsert-amiusage'});
+    } else {
+      result = await this._pgpool.query({text, values, name: 'upsert-amiusage'});
+    }
+    assert(result.rowCount === 1, 'upserting AMI usage had incorrect rowCount');
+  }
 
   /**
    * Internal method used to generate queries which have simple and condition
@@ -656,6 +680,14 @@ class State {
   async listSpotRequests(conditions = {}, client = undefined) {
     assert(typeof conditions === 'object');
     return this._listTable('spotrequests', conditions, client);
+  }
+   
+  /**
+   * Get a list of AMIs and their usage
+   */
+  async listAmiUsage(conditions = {}, client = undefined) {
+    assert(typeof conditions === 'object');
+    return this._listTable('amiusage', conditions, client);
   }
 
   /**
--- a/sql/create-db.sql
+++ b/sql/create-db.sql
@@ -89,6 +89,7 @@ CREATE TABLE IF NOT EXISTS cloudwatchlog (
 -- Amazon Machine Image (ami) usage
 CREATE TABLE IF NOT EXISTS amiusage (
   region VARCHAR(128) NOT NULL, -- ec2 region
-  id VARCHAR(128) NOT NULL,
-  lastused TIMESTAMPTZ NOT NULL,
+  id VARCHAR(128) NOT NULL, -- opaque ID per Amazon
+  lastused TIMESTAMPTZ NOT NULL, -- most recent usage
+  PRIMARY KEY(id, region)
 );
--- a/test/state_test.js
+++ b/test/state_test.js
@@ -123,8 +123,10 @@ describe('State', () => {
   it('should be empty at start of tests', async () => {
     let instances = await db.listInstances();
     let pendingSpotRequests = await db.listSpotRequests();
+    let amiUsage = await db.listAmiUsage();
     assume(instances).has.length(0);
     assume(pendingSpotRequests).has.length(0);
+    assume(amiUsage).has.length(0);
   });
 
   it('should be able to insert a spot request', async () => {
@@ -139,6 +141,14 @@ describe('State', () => {
     result = await db.listSpotRequests({region: 'us-east-1', state: 'open'});
     assume(result).has.length(0);
     result = await db.listSpotRequests({region: 'us-west-1', state: 'open'});
+    assume(result).has.length(1);
+  });
+   
+  it('should be able to filter AMI usages', async () => {
+    let result = await db.reportAmiUsage({region: defaultSR.region, id: defaultSR.id});
+    result = await db.listAmiUsage({region: 'us-east-1', id: 'r-1'});
+    assume(result).has.length(0);
+    result = await db.listAmiUsage({region: 'us-west-1', id: 'r-1'});
     assume(result).has.length(1);
   });
 
@@ -161,7 +171,7 @@ describe('State', () => {
     assume(await db.listSpotRequests()).has.length(0);
     assume(await db.listInstances()).has.length(1);
   });
-
+  
   it('should be able to upsert a spot instance, removing the spot request', async () => {
     defaultInst.srid = defaultSR.id;
 
@@ -174,7 +184,6 @@ describe('State', () => {
     assume(await db.listSpotRequests()).has.length(0);
     assume(await db.listInstances()).has.length(1);
   });
-
 
   it('should be able to update an instance', async () => {
     let firstState = 'pending';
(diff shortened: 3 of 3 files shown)


lib/state.js
```diff
@@ -569,6 +569,82 @@ class State {
 
     return result.rowCount;
   }
+   
+   /**
+   * Insert an AMI's first usage.
+   */
+  async insertAmiUsage({region, id}, client) {
```
walac: Maybe `region` and `id` should not be destructed parameters, as they are non-optional.

lib/state.js
```diff
@@ -569,6 +569,82 @@ class State {
...
+    ].join(' ');
+    let values = [region, id];
+    let name = 'insert-amiusage';
+
+    let result;
+    if (client) {
+      result = await client.query({text, values, name});
+    } else {
+      result = await this._pgpool.query({text, values, name});
+    }
```
walac: maybe better:
```javascript
insertAmiUsage(region, id, client) {
  assert(typeof region === 'string');
  assert(typeof id === 'string');

  const query = client ? client.query : this._pgpool.query;

  return query({
    text: `INSERT INTO amiusage (region, id, lastused) VALUES (${region}, ${id}, CURRENT_TIMESTAMP)`,
    name: 'insert-amiusage'
  }).then(result => assert.equal(
    result.rowCount,
    1,
    'inserting AMI usage had incorrect rowCount'
  ));
}
```

This is more compact, uses less temporary variables and it is more clear that the intention of the `if-then-else` clause just boils down to choosing which `query` method to use. Although I didn't use here, the `[].join(' ')` trick is welcome if the text lines turns out to be too long.

lib/state.js
```diff
@@ -569,6 +569,82 @@ class State {
...
+
+    let result;
+    if (client) {
+      result = await client.query({text, values, name: 'upsert-amiusage'});
+    } else {
+      result = await this._pgpool.query({text, values, name: 'upsert-amiusage'});
+    }
+
+    assert(result.rowCount === 1, 'upserting AMI usage had incorrect rowCount');
+  }
```
walac: My comments on `insertAmiUsage` apply here.

lib/state.js
```diff
@@ -569,6 +569,82 @@ class State {
...
+    
+    let result;
+    if (client) {
+      result = await client.query({text, values, name});
+    } else {
+      result = await this._pgpool.query({text, values, name});
+    }
+
+    assert(result.rowCount === 1, 'updating AMI usage had incorrect rowCount');
+  }
```
walac: Same.

lib/state.js
```diff
@@ -657,6 +733,14 @@ class State {
     assert(typeof conditions === 'object');
     return this._listTable('spotrequests', conditions, client);
   }
+   
+  /**
+   * Get a list of AMIs and their usage
+   */
+   async listAmiUsage(conditions = {}, client = undefined) {
+    assert(typeof conditions === 'object');
```
walac: This assert should be moved to `this._listTable` or knocked out if it is already there.

test/state_test.js
```diff
@@ -217,24 +234,36 @@ describe('State', () => {
...
-    assume(spotRequests).has.length(1);
-    assume(spotRequests[0]).has.property('state', secondState);
+    let newRegion = 'us-west-1';
+    await db.upsertAmiUsage({region: newRegion, id: defaultSR.id});
+    amiUsage = await db.listAmiUsage(); 
+    assume(amiUsage).has.length(1);
+    assume(amiUsage[0]).has.property('region', 'us-west-1');
   });
-
+
```
walac: nit: remove the added line in the line.

lib/state.js
jhford: I think it's probably better for consistency here if we stick to using async and await.

lib/state.js
jhford: We also have that check in _listTable, but I've tended to over-assert in this codebase, so it's probably fine to leave this in here for consistency

lib/state.js
```diff
@@ -569,6 +569,82 @@ class State {
...
+    } else {
+      result = await this._pgpool.query({text, values, name});
+    }
+    assert(result.rowCount === 1, 'inserting AMI usage had incorrect rowCount');
+  }
+   
+   /**
+   * Either insert or update an AMI's usage.  This ensures an atomic insert or
+   * update, but we never learn which one happened. Only updates the row 
+   * if the proposed time is newer than the existing time.
```
jhford: This is great functionality (checking for existing time), but we're not actually doing that in the query.  We should either start doing it or remove this from the comment.

test/state_test.js
```diff
@@ -217,24 +234,36 @@ describe('State', () => {
...
+    assume(amiUsage).has.length(1);
+    assume(amiUsage[0]).has.property('region', defaultSR.region);
 
-    defaultSR.state = secondState;
-    await db.upsertSpotRequest(defaultSR);
-    spotRequests = await db.listSpotRequests(); 
-    assume(spotRequests).has.length(1);
-    assume(spotRequests[0]).has.property('state', secondState);
+    let newRegion = 'us-west-1';
+    await db.upsertAmiUsage({region: newRegion, id: defaultSR.id});
```
jhford: This test isn't really quite on the mark.  We're defining the a `newRegion` variable, which is actually the same string as `defaultSR.region`.  If those were different strings, then this test would fail because it wouldn't result in a primary key collision and would insert a second row.

test/state_test.js
```diff
@@ -217,24 +234,36 @@ describe('State', () => {
...
+    amiUsage = await db.listAmiUsage(); 
+    assume(amiUsage).has.length(1);
+    assume(amiUsage[0]).has.property('region', 'us-west-1');
```
jhford: What we really need to be checking for here is that when we use the same `region` and `id` that the timestamp stored in the database actually changes.  Both calls of `upsertAmiUsage` (which should be renamed `reportAmiUsage`) should be passed identical values.  I can't remember the resolution of the timestamps stored here, but we might need to insert a little delay to ensure that the database ticks at least one value over between the upsert calls.

lib/state.js
```diff
@@ -569,6 +569,82 @@ class State {
...
+   */
+  async upsertAmiUsage({region, id}, client) {
+    assert(typeof region === 'string');
+    assert(typeof id === 'string');
+    
+    let text = [
```
jhford: To be more consistent with the other upserts, could we break the query up like this:
```javascript
    let text = [
      "INSERT INTO amiusage (region, id, lastused)",
      "VALUES ($1, $2, CURRENT_TIMESTAMP)",
      "ON CONFLICT (region, id) DO UPDATE SET lastused = CURRENT_TIMESTAMP"
    ].join(' ');
```

sql/create-db.sql
```diff
@@ -89,6 +89,7 @@ CREATE TABLE IF NOT EXISTS cloudwatchlog (
 -- Amazon Machine Image (ami) usage
 CREATE TABLE IF NOT EXISTS amiusage (
   region VARCHAR(128) NOT NULL, -- ec2 region
-  id VARCHAR(128) NOT NULL,
-  lastused TIMESTAMPTZ NOT NULL,
+  id VARCHAR(128) NOT NULL, -- opaque ID per Amazon
```
auni53: were these comments added intentionally?

lib/state.js
walac: I would also go with template strings.

sql/create-db.sql
rafvasq: Yes, I noticed all of the other tables also had comments so I didn't see any harm in it

jhford: over commenting is always better than under commenting, provided the comments are correct :)

test/state_test.js
```diff
@@ -161,6 +171,23 @@ describe('State', () => {
     assume(await db.listSpotRequests()).has.length(0);
     assume(await db.listInstances()).has.length(1);
   });
+   
+  it('should be able to do a spot request upsert', async () => {
```
jhford: so it looks like this unit test moved.  This is not a huge deal in any way, but it does make the review a little more complicated because i want to double check that the code that was re-added is the same.  It also makes future merges a lot easier to work on because there's less chance of a merge conflict from an unintentional change.

A couple tricks for avoiding this is `git revert -p`, which will let you selectively revert a commit.  This is basically git generating the reverse of your diff (`patch -R`), then going through each hunk and asking "want to revert this bit"?  It's a really cool feature, and `-p` is supported in lots of other places too!

test/state_test.js
```diff
@@ -218,23 +244,26 @@ describe('State', () => {
...
-    let spotRequests = await db.listSpotRequests(); 
-    assume(spotRequests).has.length(1);
-    assume(spotRequests[0]).has.property('state', firstState);
-
-    defaultSR.state = secondState;
-    await db.upsertSpotRequest(defaultSR);
-    spotRequests = await db.listSpotRequests(); 
-    assume(spotRequests).has.length(1);
-    assume(spotRequests[0]).has.property('state', secondState);
+    await db.reportAmiUsage({region: defaultSR.region, id: defaultSR.id});
```
jhford: Let's use the ImageId instead of a spot request Id.  This isn't important now (it's just a string) but when things go wrong in the future, it'll be confusing to see "sr-1234" when we're expecting "ami-1234".  The other thing is that we might start doing checks in the reportAmiUsage method to make sure that the imageId actually matches a pattern.  Not important, but it's also a pretty easy fix, using `defaultSR.imageId` instead of `.id`.

test/state_test.js
```diff
@@ -218,23 +244,26 @@ describe('State', () => {
...
+    let lastUse = amiUsage[0].lastused;
+    
+    await db.reportAmiUsage({region: defaultSR.region, id: defaultSR.id});
+    let updatedAmiUsage = await db.listAmiUsage();
+    assume(amiUsage).has.length(1);
+    assume(amiUsage[0]).has.property('region', defaultSR.region);
+    assume(amiUsage[0]).has.property('id', defaultSR.id);
+    let updatedUse = updatedAmiUsage[0].lastused;
+    
+    assume(lastUse < updatedUse).true();
```
jhford: Great!

test/state_test.js
rafvasq: thanks for the info @jhford , will play around with this and fix today

lib/state.js
```diff
@@ -657,6 +683,14 @@ class State {
     assert(typeof conditions === 'object');
     return this._listTable('spotrequests', conditions, client);
   }
+   
+  /**
+   * Get a list of AMIs and their usage
+   */
+   async listAmiUsage(conditions = {}, client = undefined) {
```
auni53: There's an extra space before the `async` here.

test/state_test.js
rafvasq: @jhford, with the help of @walac I managed to rebase and do some squashing. I think I even managed to leave this unit test in the same place as it was before, let me know what you think

jhford:
Looks good to me!

jhford:
rebased and pushed to master branch with 12a8aac60ad3953022dd884e35c29cc3cb950034

Thanks!

PR #36, opened 2017-10-30
Add ami-existence poller

auni53:
"Implement a function to check for existence of an AMI image, including unit tests which do not depend on real AWS requests"

### DIFF
--- a/lib/ami-existence-poller.js
+++ b/lib/ami-existence-poller.js
@@ -1,0 +1,81 @@
+const assert = require('assert');
+const {runAWSRequest} = require('./aws-request');
+const Iterate = require('taskcluster-lib-iterate');
+
+const log = require('./log');
+
+const BATCH_SIZE = 100;
+
+class AmiExistencePoller {
+
+  constructor({ec2, runaws = runAWSRequest, amiPollDelay = 25, state}) {
+    assert(typeof ec2 === 'object');
+    assert(typeof state === 'object');
+
+    this.ec2 = ec2;
+    this.state = state;
+    this.runaws = runaws;
+    this.amis = [];
+
+    this.iterator = new Iterate({
+      maxIterationTime: 1000 * 60 * 15,
+      watchDog: 1000 * 60 * 15,
+      maxFailures: 10,
+      waitTime: amiPollDelay,
+      handler: async(watchdog) => {
+        try {
+          await this.poll();
+        } catch (err) {
+          console.dir(err.stack || err);
+          monitor.reportError(err, 'Error polling pricing data');
+        }
+      },
+    });
+  };
+
+  start() {
+    return this.iterator.start();
+  }
+
+  stop() {
+    return this.iterator.stop();
+  }
+
+  async loadAmis(amis) {
+    this.amis = amis ? amis : (await this.state.listAmiUsage()).map(o => o.id);
+  }
+
+  async poll() {
+    assert(Array.isArray(this.amis), 'this.amis is not an array');
+    let result = [];
+
+    let i = 0;
+    while (i < this.amis.length) {
+      result = [...result, ...await this.pollBatch(this.amis.slice(i, BATCH_SIZE))];
+      i += BATCH_SIZE;
+    }
+
+    return result;
+  }
+
+  /**
+   * Return a list of amiIds not available in EC2
+   */
+  async pollBatch(amiIds) {
+    assert(Array.isArray(amiIds), 'amiIds is not an array');
+    amiIds.forEach(amiId => {
+      assert(typeof amiId === 'string', `ami ${amiId} is not a string, rather type ${typeof amiId}`);
+    });
+
+    const {Images} = await this.runaws(this.ec2, 'describeImages', {
+      ImageIds: [amiIds],
+    });
+    assert(Array.isArray(Images), 'Invalid aws response');
+
+    const imagesMap = new Map(Images.map(i => [i.ImageId, i.State]));
+
+    return amiIds.filter(ami => !(imagesMap.has(ami) && imagesMap.get(ami) === 'available'));
+  }
+}
+
+module.exports = {AmiExistencePoller};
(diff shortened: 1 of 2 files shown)


auni53:
@jhford the CI build is failing because it can't find the `amiusage` table. Not sure if it's a git issue or a code issue.

lib/ami-existence-poller.js
```diff
@@ -0,0 +1,56 @@
...
+    this.runaws = runaws;
+
+    this.iterator = new Iterate({
+      maxIterationTime: 1000 * 60 * 15,
+      watchDog: 1000 * 60 * 15,
+      maxFailures: 10,
+      waitTime: amiPollDelay,
+      handler: async (watchdog) => {
+        try {
+          await this.poll();
```
jhford: It looks like this is missing an argument to the function.  That seems right, since we want to do batches of ami ids at the same time, but as implemented, this is going to throw an assertion failure on the first call.

lib/ami-existence-poller.js
```diff
@@ -0,0 +1,56 @@
...
+    return this.iterator.stop();
+  }
+
+  async poll(amiId) {
+    assert(typeof amiId === 'string', 'ami is not string, rather a ' + typeof ami);
+    const amisInDatabase = await this.state.listAmiUsage({id: amiId});
+    assert(amisInDatabase.length, 'ami not in database');
+
+    const {Images} = await this.runaws(this.ec2, 'describeImages', {
+      ImageIds: [amiId],
```
jhford: So this will definitely work, but it will require one EC2 Api call for each AMI ID that we'd like to look up.  It would be great if this function instead took either a list of Ami ids, or none and loaded them from the DB, then checked on batches of 100 per EC2 API call.

test/ami-existence-poll_test.js
```diff
@@ -0,0 +1,149 @@
...
+const sinon = require('sinon');
+const main = require('../lib/main');
+const assume = require('assume');
+
+describe('AMI Existence Poller', () => {
+  let state;
+  let sandbox = sinon.sandbox.create();
+  let defaultSR;
+
+  before(async () => {
```
jhford:
```javascript
 10   before(async() => {
 11     db = await main('state', {profile: 'test', process: 'test'});
 12     await db._runScript('drop-db.sql');
 13     await db._runScript('create-db.sql');
 14   });
```

I think the issues you're having with the table not existing is that the create-db.sql script isn't getting run before the tests run.  Since this file alphabetically sorts early, it probably is being run after the database is set up by another unit tests.  This highlights though, that the other DB tests should be dropping the database in their `after()` calls, to ensure consistent behaviour.

lib/ami-existence-poller.js
```diff
@@ -0,0 +1,56 @@
+const assert = require('assert');
+const {runAWSRequest} = require('./aws-request');
+const Iterate = require('taskcluster-lib-iterate');
+
+const log = require('./log');
+
+const BATCH_SIZE = 100;
```
jhford: We're not using this anywhere, but we should be.

test/ami-existence-poll_test.js
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  describe('constructor', () => {
+    it('succeeds with valid args', () => {
+      const poller = new AmiExistencePoller({ec2: {}, state, runaws: () => {}});
+    });
+  });
+
+  describe('polling', () => {
+
```
rafvasq: nit: extra line

test/ami-existence-poll_test.js
```diff
@@ -0,0 +1,149 @@
...
+      const poller = new AmiExistencePoller({ec2: {}, state, runaws: mock});
+      try {
+        await poller.poll(defaultSR.imageId);
+        return Promise.reject(Error('Line should not be reached'));
+      } catch (e) { }
+    });
+
+    describe('with the ami present in the database', () => {
+
```
rafvasq: nit: blank line

test/ami-existence-poll_test.js
```diff
@@ -0,0 +1,252 @@
...
+          State: 'inavailable',
+        }],
+      }));
+
+      const poller = new AmiExistencePoller({ec2: {}, state, runaws: mock});
+      await poller.loadAmis([defaultSR.imageId]);
+      assume(await poller.poll()).eql([defaultSR.imageId]);
+    });
+
+    describe('batches correctly', async() => {
```
jhford: nice tests!

jhford:
I just noticed that the issue is still there about the before() call missing the db initialization.  The code from https://github.com/taskcluster/ec2-manager/pull/36#discussion_r147712817 needs to be in this diff, so that the database is created for the tests.

auni53:
@jhford is this good to merge in?

lib/ami-existence-poller.js
```diff
@@ -0,0 +1,82 @@
...
+
+  async loadAmis(amis) {
+    this.amis = amis ? amis : (await this.state.listAmiUsage()).map(o => o.id);
+  }
+
+  async poll() {
+    assert(Array.isArray(this.amis), 'this.amis is not an array');
+    let result = [];
+
+    // parallelize this with Promise.all?
```
jhford: We could, but we're trying to minimize our API request rate, and since this is not involved in responsiveness of the API, let's leave them sequential.

test/ami-existence-poll_test.js
```diff
@@ -0,0 +1,254 @@
...
+const assume = require('assume');
+
+describe('AMI Existence Poller', () => {
+  let state;
+  let sandbox = sinon.sandbox.create();
+  let defaultSR;
+
+  before(async() => {
+    state = await main('state', {profile: 'test', process: 'test'});
+    db = await main('state', {profile: 'test', process: 'test'});
```
jhford: no need to create a second copy of the state.  let's pick either `db` or `state` and drop the other.

PR #37, opened 2017-10-30
Drop the SQL tables after each suite that uses it.

jhford:
Thanks auni53 for highlighting this issue.  There's a bit of a problem
here where the order of test files will have a profound effect on
whether or not the tests pass or not.  Each test file should delete
existing tables, create them and then delete them when they're done.  If
there was a way to do a before()/after() that spans all the tests, this
would be a great place to run a database drop and create.


jhford:
This should help us with issues around the SQL table creation/dropping order.

PR #39, opened 2017-11-06
Add endpoint for /internal/amiusage

rafvasq:
Implemented an endpoint “GET /internal/ami-usage” which lists out all of the AMIs and their usage and included unit test

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -577,6 +577,22 @@ api.declare({
 });
 
 /**
+ * List AMIs and their usage
+ */
+api.declare({
+  method: 'get',
+  route: '/internal/ami-usage',
+  name: 'amiUsage',
+  title: 'See the list of AMIs and their usage',
+  stability: API.stability.experimental,
+  scopes: [['ec2-manager:internals']],
+  description: 'Lists out all of the AMIs and their usage',
+}, async function(req, res) {
+  let amiUsage = await this.state.listAmiUsage();
+  return res.reply(amiUsage);
+});
+
+/**
  * Show stats on the Database Pool
  */
 api.declare({
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -424,5 +424,16 @@ describe('Api', () => {
       assume(usw2.values).has.lengthOf(1);
       assume(usw2.values[0]).equals('r-1234');
     });
+  
+    it('should list AMI usage', async() => {
+      await state.reportAmiUsage({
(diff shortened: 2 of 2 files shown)


lib/api.js
```diff
@@ -577,6 +577,22 @@ api.declare({
...
+api.declare({
+  method: 'get',
+  route: '/internal/ami-usage',
+  name: 'amiUsage',
+  title: 'See the list of AMIs and their usage',
+  stability: API.stability.experimental,
+  scopes: [['ec2-manager:internals']],
+  description: 'Lists out all of the AMIs and their usage',
+}, async function(req, res) {
+  let amiUsage = await this.state.listAmiUsage();
```
walac: It has been a standard in this project, so it is okay, but bear in mind that in general, variables should be `const` by default and we only use `let` for cases where we have good reasons to modify them (like the control variable in a `for` loop).

lib/api.js
```diff
@@ -577,6 +577,22 @@ api.declare({
 });
 
 /**
+ * List AMIs and their usage
+ */
+api.declare({
+  method: 'get',
+  route: '/internal/ami-usage',
+  name: 'amiUsage',
+  title: 'See the list of AMIs and their usage',
```
walac: nit: s/See/Show/

lib/api.js
```diff
@@ -577,6 +577,22 @@ api.declare({
...
+  stability: API.stability.experimental,
+  scopes: [['ec2-manager:internals']],
```
walac: I am not very much familiar with the semantics of the service. Therefore I will leave review of scopes management to @jhford

lib/api.js
rafvasq: Good to know, thanks for the info!

lib/api.js
rafvasq: I would go with "show" as well but other GET methods use "See the lists ..." so I decided to be consistent

lib/api.js
walac: Good point.

lib/api.js
jhford: Let's stick with consistency.

lib/api.js
jhford: This is an endpoint that's only useful to the taskcluster team, and as a /internals/ endpoint, i'd like to keep it locked down like this.  This is the right scope.

PR #43, opened 2017-11-08
Allow deletion of worker type resources with a worker-type scope

jhford:
Instead of forcing a specific instanceId/requestId/region scope, let's
say that if the desired instance/request is managed by this process and
has a worker type for which the appropriate scope has been granted, you
can kill/cancel the instance/request

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -409,7 +409,10 @@ api.declare({
   name: 'terminateInstance',
   title: 'Terminate an instance',
   stability: API.stability.experimental,
-  scopes: [['ec2-manager:manage-instances:<region>:<instanceId>']],
+  scopes: [
+    ['ec2-manager:manage-instances:<region>:<instanceId>'],
+    ['ec2-manager:manage-resources:<workerType>'],
+  ],
   description: [
     'Terminate an instance in a specified region',
   ].join(' '),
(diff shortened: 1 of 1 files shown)


jhford:
cc @helfi92

lib/api.js
```diff
@@ -409,15 +409,41 @@ api.declare({
   name: 'terminateInstance',
   title: 'Terminate an instance',
   stability: API.stability.experimental,
-  scopes: [['ec2-manager:manage-instances:<region>:<instanceId>']],
+  scopes: [
+    ['ec2-manager:manage-instances:<region>:<instanceId>'],
+    ['ec2-manager:manage-resources:<workerType>'],
```
jonasfj: I would suggest something like `....:<provisionerId>/<workerType>/<workerGroup>/<workerId>` notably because we can just give out the scope: `....:<provisionerId>/<workerType>/*`, and then you can manage any instance under said `workerType`

I'll admit I don't know when you would want a `workerId` specific scope, but who knows...

Note: `workerType` is only expected to be unique wrt. to `provisionerId`. So if you don't including `<provisionerId>` you might be in trouble. Granted this service is sort of internal to aws-provisioner-v1, or ??? at-least that's how I understand it...

I will say I'm a little confused about the boundaries, since this service seems to somewhat deal with `workerType` but also use a lot of EC2 specific terms that aren't generic, like `instanceId`. Anyways, that's all waay off topic here :) hehe

lib/api.js
```diff
@@ -409,15 +409,41 @@ api.declare({
   name: 'terminateInstance',
   title: 'Terminate an instance',
   stability: API.stability.experimental,
-  scopes: [['ec2-manager:manage-instances:<region>:<instanceId>']],
+  scopes: [
+    ['ec2-manager:manage-instances:<region>:<instanceId>'],
```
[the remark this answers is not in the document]
jhford: Good idea!  I don't want to risk changing all the scopes right now, so for the time being I'll just use the :subresource notation to avoid that risk.

lib/api.js
jhford: Yep, that was an oversight while implementing originally.  At the present time, there's really only one aws provisioner, so it's not a high priority.  I'll add that level at some point, but since we need different credentials for each ec2-account, it's not trivial to add this support.

PR #45, opened 2017-11-29
Store the EBS totals in Postgres and fetch these totals in an API call

SamanthaYu:
## Store the EBS totals in Postgres
We decided to store the EBS totals in Postgres. In the table `ebsusage`, the columns are `region`, `volumetype`, `state`, `totalcount`, `totalgb`, and `touched` with the primary key a combination of `region`, `volumetype`, and `state`. When `reportEbsUsage()` gets called, it inserts the totals into their respective columns. If there's already an existing record with that primary key, then that record's `totalcount` and `totalgb` gets replaced with the reported `totalcount` and `totalgb` respectively, and `touched` is set to the current timestamp.

## Fetch these totals in an API call
The internal API call, `ebsUsage()`, returns the entire `ebsusage` table. In other words, it returns a list of objects in the form:
```
{
   region: string,
   volumetype: string,
   state: string,
   totalcount: string,
   totalgb: string,
   touched: timestamp,
}
```

## Updated documentation for AMI API call
Because the API calls for EBS usage and AMI usage are similar, I updated the AMI API call's for consistency with the EBS API call and for clarity of what values get returned:
```
/**
 * List AMIs and their usage by returning a list of objects in the form:
 * {
 *   region: string,
 *   volumetype: string,
 *   lastused: timestamp,
 * }
 */
```
@rafvasq and @auni53 What do you think? Is this comment accurate?

## Other comments
@rstiyer and I decided to combine our two parts into this PR. Since @rstiyer worked on the database side and I worked on the EBS API call, the tests for each of our parts would rely on each other so it would be hard to separate them completely.

@djmitche and @imbstack Can you review this PR when you get the chance?


sql/create-db.sql
```diff
@@ -93,3 +93,16 @@ CREATE TABLE IF NOT EXISTS amiusage (
...
 );
+
+-- Elastic Block Store (EBS) Volumes uasge
+-- We are only interested in storing the total number of volumes and the total number
+-- of GBs per type/region from the most recent housekeeping sweep.
+CREATE TABLE IF NOT EXISTS ebsusage (
+  region VARCHAR(128) NOT NULL, -- ec2 region
+  volumetype VARCHAR(128) NOT NULL,  -- volume type (e.g gp2, st1)
+  state VARCHAR(128) NOT NULL CHECK (state in ('active', 'unused')), -- state of aggregated volumes
+  totalcount VARCHAR(128) NOT NULL, -- number of volumes in region, of volumetype and state
```
SamanthaYu: @rstiyer Why is `totalcount` and `totalgb` stored as a string instead of an integer?

lib/api.js
```diff
@@ -638,7 +638,12 @@ api.declare({
 });
 
 /**
- * List AMIs and their usage
+ * List AMIs and their usage by returning a list of objects in the form:
+ * {
+ *   region: string,
+ *   volumetype: string,
+ *   lastused: timestamp,
+ * }
```
djmitche: Thanks for this!  However, docs for API methods should be in the `description` field a few lines below.  That eventually makes its way to the documentation site so that others know how to call these API methods.  It's in markdown format.

lib/api.js
```diff
@@ -653,6 +658,30 @@ api.declare({
...
+/**
+ * Lists current EBS volume usage by returning a list of objects in the form:
+ * {
+ *  region: string,
+ *  volumetype: string,
+ *  state: string,
+ *  totalcount: string,
+ *  totalgb: string,
+ *  touched: timestamp,
+ * }
```
djmitche: You might also mention that there is one row per region+volumetype+state, and that `touched` is the last time that information was updated.

lib/housekeeping.js
```diff
@@ -408,7 +437,7 @@ class HouseKeeper {
 
     // Calculate totals and update statum series after all regions' volume data
     // has been accumulated.
-    this._handleVolumeData(volumesByRegion);
+    await this._handleVolumeData(volumesByRegion);
```
djmitche: Nice job adding async/await throughout this call chain!

lib/housekeeping.js
```diff
@@ -136,20 +141,44 @@ class HouseKeeper {
         volumesMonitor.measure('${region}.type:${type}.count', combinedCount);
         totalGbUsage += combinedGbUsage;
         totalCount += combinedCount;
-      });
-    });
-    
+        
+        let transaction = await this.state.beginTransaction();
+        try {
```
djmitche: If I remember correctly, we want to ensure that rows don't "stick around" in this table if they are no longer being updated -- for example, if we stop monitoring a region.  I think the easiest way to accomplish that is simply to delete all rows from the table at the beginning of the transaction.  In that case, you won't need the upsert, either -- an insert will do, since you know the table is otherwise empty.

sql/create-db.sql
```diff
@@ -93,3 +93,16 @@ CREATE TABLE IF NOT EXISTS amiusage (
...
   PRIMARY KEY(id, region)
 );
+
+-- Elastic Block Store (EBS) Volumes uasge
+-- We are only interested in storing the total number of volumes and the total number
+-- of GBs per type/region from the most recent housekeeping sweep.
+CREATE TABLE IF NOT EXISTS ebsusage (
+  region VARCHAR(128) NOT NULL, -- ec2 region
+  volumetype VARCHAR(128) NOT NULL,  -- volume type (e.g gp2, st1)
+  state VARCHAR(128) NOT NULL CHECK (state in ('active', 'unused')), -- state of aggregated volumes
```
djmitche: if these are active/unused, they don't need 128 bytes :)

test/housekeeping_test.js
```diff
@@ -101,10 +101,6 @@ describe('House Keeper', () => {
     houseKeeperMock = sandbox.mock(houseKeeper);
   });
 
-  after(async() => {
-    await state._runScript('drop-db.sql');
-  });
-
   afterEach(() => {
```
djmitche: why remove this drop call?

lib/api.js
SamanthaYu: I just realized that there's this comment on lines 598 to 602 saying:
```
/*    NOTE:  ALL FOLLOWING METHODS ARE INTERNAL ONLY AND ARE NOT             */
/*           INTENDED FOR GENERAL USAGE.  AS SUCH THEY ARE ALL               */
/*           CONSIDERED TO BE EXPERIMENTAL, DO NOT HAVE ANY                  */
/*           SCHEMA DESCRIPTIONS AND ARE INTENDED TO BE CHANGED              */
/*           WITHOUT ANY NOTICE.                                             */
```
Should we still specify the schema for AMI and EBS usage then?

Also, where is the internal API documentation site?

djmitche: There isn't an internal API documentation site :)

Schema descriptions are formal, verifiable JSON-schema entries.  As the comment suggests, we don't need to add those.  But I think it's just neighborly to describe (in the description) what the results look like, to avoid misunderstandings and misinterpretation of the results.

sql/create-db.sql
rstiyer: Good point! I will make those much smaller :)

lib/api.js
```diff
@@ -652,14 +652,25 @@ api.declare({
...
+    [
+      'List AMIs and their usage by returning a list of objects in the form:',
+      '{',
+      [
+        'region: string',
+        'volumetype: string',
+        'lastused: timestamp',
+      ].join('\n\t'),
+      '}',
+    ].join('\n'),
```
djmitche: :heart:

sql/create-db.sql
rstiyer: That was a silly thing to do, you're right! I will make those numeric.

lib/api.js
auni53: Would this be neater with the `` query strings?

lib/housekeeping.js
```diff
@@ -123,9 +124,13 @@ class HouseKeeper {
         assert(active.hasOwnProperty('count'));
         assert(unused.hasOwnProperty('gb'));
         assert(unused.hasOwnProperty('count'));
-
+
```
auni53: trailing space?

sql/create-db.sql
auni53: @jhford seemed to imply that's unnecessary?

https://github.com/taskcluster/ec2-manager/pull/29#discussion_r142249195

lib/housekeeping.js
SamanthaYu: Probably but I wasn't sure about the whitespace. I'll look more into this in a few days.

I'm blaming Atom for the trailing space. It keeps creating random indents. There's probably a setting in Atom that I could turn off.

auni53: https://en.wikipedia.org/wiki/Vim_(text_editor)

djmitche:
I think -- please correct me -- the current state is that this is waiting for an update from rachel to overwrite the DB in a single transaction.  With that change (and with the drop-database call restored in test_housekeeping.js) this should be good to go.

lib/housekeeping.js
```diff
@@ -142,35 +146,38 @@ class HouseKeeper {
...
+      // Submit row data to ebsusage table
+      let transaction = await this.state.beginTransaction();
+      try {
+        await this.state.reportEbsUsage(rowData); //, transaction);
+        await this.state.commitTransaction(transaction);
+      } catch (err) {
+        await this.state.rollbackTransaction(transaction);
+        log.error(err, 'Error reporting EBS usage');
+      }
+    }
```
djmitche: It looks like this wraps the `reportEbsUsage` call in a transaction, but `reportEbsUsage` also uses its own transaction.  In fact since the `transaction` arg is commented out here, I think this transaction isn't used at all and could be removed (making this a one-line call to `reportEbsUsage`).

lib/state.js
```diff
@@ -872,32 +872,46 @@ class State {
...
-    assert(typeof region === 'string');
-    assert(typeof volumetype === 'string');
-    assert(typeof state === 'string');
-    assert(typeof totalcount === 'number');
-    assert(typeof totalgb === 'number');
+  async reportEbsUsage(rowData) { //, client) {
+    assert(typeof rowData === 'object');
+    assert(Array.isArray(rowData));
+
+    let transaction = await this.beginTransaction();
```
djmitche: This is good, but needs a try/catch to roll the transaction back if something goes wrong.

lib/state.js
```diff
@@ -872,32 +872,46 @@ class State {
...
+    let text = 'TRUNCATE TABLE ebsusage;';
+    
     if (client) {
-      result = await client.query({text, values, name});
+      await client.query(text);
     } else {
-      result = await this._pgpool.query({text, values, name});
+      // We should never arrive in this case, since we clear the database and insert
+      // new values in the same transaction.
+      throw new Error('Client should not be undefined in clearEbsUsage');
```
djmitche: It might be easier to just write `await transaction.query('TRUNCATE TABLE ebsusage');` in `reportEbsUsage`?

lib/housekeeping.js
rstiyer: Whoa, this is interesting. None of the tests fail for me locally :confused:

rstiyer:
I added a check within the function `reportEbsUsage(rowData)` that does nothing if `rowData` is an empty array, but I wanted to confirm if everyone thinks this makes sense. The alternative would be to assume that the empty list is correct, so just clear the table and insert nothing.

djmitche:
I think if there's no row data then we're provisioning in 0 regions, which doesn't seem right.  So I think it's fine as it is.

PR #48, opened 2017-12-20
Spring cleaning fall 2017

djmitche:


lib/main.js
```diff
@@ -378,7 +378,6 @@ if (!module.parent) {
     log.fatal({err}, 'Unhandled Promise Rejection!');
     throw err;
   });
-  require('source-map-support').install();
```
jhford: thanks!

package.json
```diff
@@ -39,12 +41,9 @@
   },
   "devDependencies": {
     "assume": "^1.5.1",
-    "babel-eslint": "^8.0.1",
     "eslint": "^4.9.0",
-    "eslint-config-taskcluster": "^2.0.0",
-    "eslint-plugin-taskcluster": "^1.0.2",
+    "eslint-config-taskcluster": "^3.0.0",
```
jhford: oh neat! we declared all the deps as deps of this package?  great!

Procfile
```diff
@@ -1 +1 @@
-web: node lib/main start | ./node_modules/.bin/bunyan -o short
```
jhford: Sorry, I'd like to leave this as `lib/`.  This tool was the first to use babel-less node again, and has used `lib/` from day one.  Best practises should have been written after examining prior art like this repository.

As for the reason, this code is not source for a compiler, rather it's executable code.

src/main.js
```diff
@@ -82,6 +83,21 @@ let load = loader({
     },
   },
 
+  docs: {
```
jhford: do we have something here to ensure that docs are only published for those environments which have requested publishing in their config.yml entry?

package.json
```diff
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "aws-sdk": "^2.172.0",
-    "bluebird": "^3.5.0",
+    "bluebird": "^3.5.1",
```
jhford: I'm pretty sure we're not actually using bluebird at all...  It can probably be deleted.

config.yml
```diff
@@ -56,7 +55,7 @@ production:
     env: production
     forceSSL: true
     trustProxy: true
-    baseUrl: 'https://ec2-manager.herokuapp.com/v1'
```
jhford: I was under the impression that the base url was supposed to have the `/v1` in it

src/main.js
```diff
@@ -119,7 +119,7 @@ let load = loader({
           instancePubKey: cfg.app.instancePubKey,
           state: state,
           regions: cfg.app.regions,
-          apiBaseUrl: cfg.server.baseUrl,
+          apiBaseUrl: cfg.server.publicUrl + '/v1',
```
jhford: ahh, you're adding it here.  Let's leave this as a hard coded value in the config.

Procfile
djmitche: I find these little differences trip me up repeatedly when working across multiple services, for example when upgrading a widely-used TC library.  Remembering that ec2-manager is "different" is just a little cognitive drag.

It will take quite a bit of time to shift every other service over to `lib/`, rather than shifting this one to `src/`.  Is it worth spending that time for three letters?

src/main.js
djmitche: This is bringing the approach into line with [taskcluster-queue](https://github.com/taskcluster/taskcluster-queue/blob/master/src/main.js#L408) and other services.  That service doesn't send the API base URL as context, but does include it as an argument to `v1.setup(..)`.

Procfile
djmitche: We agreed to just leave these different.

src/main.js
djmitche: The default is
```
    publish: process.env.NODE_ENV == 'production',
```

so this shouldn't need any special handling.

djmitche:
I dropped the lib/ -> src/ move.

I think that handling `publicUrl` consistently will be important when we are deploying all of the services using taskcluster-standalone, so I'd like to keep that in.

djmitche:
Conversation yesterday indicated @jhford is manually rebasing this and will land.

src/main.js
jhford: nope, NODE_ENV is set to production even for staging apps.  We should also not be reading the NODE_ENV variable outside of `lib/main.js`

djmitche: OK, then we should pass an explicit `publish` to this argument, based on a `config.yml` value.

djmitche:
I don't know if this ever landed, but it's too far bitrotted to recover now..

PR #49, opened 2018-01-08
Implement new spot request model

jhford:
This PR implements the changes needed for us to use the new spot model.  In this model, spot requests are made by using the on-demand runInstances endpoint with an extra parameter which provides information about how the spot request should be handled.  No attempt has been made to have a clean cutover for existing spot request objects.  We'll just let them exist outside of our real view as they should all disappear within a short time frame and not have too big an impact.  Existing instances will continue to be tracked, so we shouldn't lose track of them.

There are some SQL schema changes.  I've written a script (sql/deploy-new-spot-changes.sql) which I've tested to run and edit the database as expected.

The major changes:

1. Removal of all handling for spot instance requests and for making calls to requestSpotInstance(s)
2. Implementation of runInstances, with logic for handling spot requests
3. Resulting simplification of transaction logic for instance state manipulation
4. Temporary removal of EBS volume monitoring -- it was broken and I couldn't figure out what was going wrong.  Will be readded at a later date.
5. Case preservation of Postgres column names -- should make it easier to do queries


jhford:
cc @walac: Here's a preview of the branch I'm working against right now.  Feel free to take a look at the review.

jhford:
Note that this is a pretty extensive patch!

lib/api.js
```diff
@@ -74,47 +74,73 @@ api.declare({
...
+
+    if (RequestType === 'spot') {
+      LaunchInfo.InstanceMarketOptions = {
+        MarketType: 'spot',
+        SpotOptions: {
+          SpotInstanceType: 'one-time',
+        },
+      };
+
+      // using typeof to account for a 0.0 value... I guess no one would use a
```
walac: I think I am losing some context, but I didn't understand how the `typeof` handles the 0.0 case...

lib/main.js
```diff
@@ -383,6 +371,7 @@ if (!module.parent) {
     profile: process.env.NODE_ENV || 'development',
     process: process.argv[2],
   }).catch(err => {
+    console.dir(err);
```
walac: wouldn't be better to use `console.log(err.stack || err)`?

jhford:
@imbstack hey, do you mind taking a look at the patch while I work on the provisioner patch?

test/state_test.js
```diff
@@ -79,7 +79,7 @@ describe('State', () => {
 
     it('one flat condition', () => {
       let expected = {
-        text: 'SELECT * FROM junk WHERE junk.a = $1;',
+        text: 'SELECT * FROM junk WHERE junk."a" = $1;',
```
walac: Just out of curiosity: why the quotation marks?

lib/api.js
```diff
@@ -74,47 +74,73 @@ api.declare({
...
-      SpotPrice: SpotPrice.toString(),
-      ClientToken,
-      LaunchSpecification,
-      InstanceCount: 1,
-    };
+    // lsChecker has already ensured that these keys aren't set so no need to
+    // duplicate the effort
+    LaunchInfo.ClientToken = ClientToken;
+    // Once we have a different solution for security tokens, we can make this
+    // configurable
```
imbstack: \o/

lib/api.js
```diff
@@ -157,14 +157,16 @@ api.declare({
     assert(result.Instances.length === 1);
 
     let [instance] = result.Instances;
+    
+    console.dir(instance);
```
imbstack: Do we want to leave this in?

lib/api.js
jhford:
```js
$ node
> if (0.0) { console.log('truthy') }
undefined
> if (typeof 0.0 === 'number') { console.log('truthy') }
truthy
undefined
> 
```
It's being pedantic, but I want to make sure that if a value of 0 is passed, which is a falsy value, it still triggers this value to be added to the request.  I know a bid of 0 is not a great thing, but I'm worried about the case where some automated process starts accidentally calculating bids of 0$.

test/state_test.js
jhford: In this specific case it's overkill, but they're used for case-sensitivity.  It was easier to just quote everything even if unneeded than to determine if there were caps in the column name and only quote then.  Really though, in both cases, we should be using parameterized condition names.

jhford:
This set of patches is ready for final review.  Basically, I removed the support for spot requests anywhere.  Because we have an atomic call for creating instances, we can know from our database which instances belong to us.  This means that we can avoid doing a lot of describeInstances API calls, and the associated performance and reliability hits.

In the new system, we don't actually need to make any describe* calls as part of tracking instance state.

There's also preliminary support for tracking the creation of terminations.  This table currently does not store codes or reasons for terminations, but in the future, we'll be using the State.findTerminationsToPoll() method to do batched polling of instances using the describeInstances API to determine termination reason so that we can build the biaser on that.

Please give this a good look as it's changing a lot of fundementals of how EC2-Manager works.  I'm also happy to help with any questions about it, or hop on a vidyo chat if that makes things easier.

lib/cloud-watch-event-listener.js
```diff
@@ -168,14 +159,41 @@ class CloudWatchEventListener extends events.EventEmitter {
         metadataSource: 'db',
       };
       
+      // Note that since we're now inserting instances directly into the
+      // database now, we're always going to have a pending state in the db for
+      // managed instances.  This means that most 'pending' state messages will
+      // be ignored.
+      if (dbResponse.lastEvent && dbResponse.lastEvent === generated) {
+        log.warn('whoa!');
```
imbstack: Should we send a warning to sentry instead?

lib/cloud-watch-event-listener.js
```diff
@@ -168,14 +159,41 @@ class CloudWatchEventListener extends events.EventEmitter {
...
+            await this.state.upsertTermination({
+              id: dbResponse.id,
+              workerType: dbResponse.workerType,
+              region: dbResponse.region,
+              az: dbResponse.az,
+              instanceType: dbResponse.instanceType,
+              imageId: dbResponse.instanceType,
+              launched: dbResponse.launched,
+              terminated: generated,
+              lastEvent: generated,
```
imbstack: I know `generated` isn't a new term here, but I keep finding it confusing as to what it is. Maybe indicate that it is a date or something?

lib/cloud-watch-event-listener.js
```diff
@@ -194,81 +212,15 @@ class CloudWatchEventListener extends events.EventEmitter {
...
-        if (workerType && missingTags(instance)) {
-          await this.tagger.tagResources({
-            ids: [id],
-            workerType: workerType,
-            region: region,
-          });
-        }
-        let opts = {
-          workerType,
+        log.info({
```
imbstack: If this is something we want to watch, should it be a sentry warning or even just a sentry info?

lib/cloud-watch-event-listener.js
jhford: It's not strictly an error, not really even something we monitor, more just something I'd be interested to see if it ever happens.

lib/cloud-watch-event-listener.js
jhford: it's the Date that the message from cloudwatch was generated.  I could rename it, but then we'd have a really long date.  There is a comment explaining that, yes, the name sucks, but explains what it is.
```js
    // Not a great name, but this variable is the time that the CloudWatch
    // Event message was generated by the API, not the time at which we
    // received it
    let generated = new Date(body.time);
```

test/cloud-watch-event-listener_test.js
```diff
@@ -153,14 +136,17 @@ describe('Cloud Watch Event Listener', () => {
...
+    assume(instances[0]).has.property('lastEvent');
+    assume(instances[0].lastEvent.getTime()).equals(runningTimestamp.getTime());
     assume(instances[0].state).equals('running');
 
     assume(instances).lengthOf(1);
   });  
 
   it('should handle out of order delivery', async() => {
+    let mock = sandbox.stub(listener, 'runaws');
+    mock.onFirstCall().throws(new Error('shouldnt talk to ec2 api'));
```
walac: nit: shouldn't

jhford:
this landed!

PR #50, opened 2018-01-15
Bug 1428068: Handle EC2 prices endpoint pagination

walac:
The DescribeSpotPriceHistory endpoint returns a field call NextToken
that, if not null, indicates there are additional data to fetch.

On next call, NextToken should be an input parameters to tell the server
we are fetching the remaining data.

Notice it had to move DescribeAvailabilityZones out of the loop because
it doesn't do pagination.

### DIFF
--- a/lib/pricing-poller.js
+++ b/lib/pricing-poller.js
@@ -135,44 +135,57 @@ class PricingPoller {
 
     for (let region of this.regions) {
       let ec2 = this.ec2[region];
-      let [azResult, priceResult] = await Promise.all([
-        this.runaws(ec2, 'describeAvailabilityZones', {
-          Filters: [{
-            Name: 'state',
-            Values: ['available'],
-          }],
-        }),
-        this.runaws(ec2, 'describeSpotPriceHistory', {
+      let nextToken = undefined;
+      let azResult = await this.runaws(ec2, 'describeAvailabilityZones', {
+        Filters: [{
+          Name: 'state',
+          Values: ['available'],
+        }],
+      });
+      let zones = azResult.AvailabilityZones.map(x => x.ZoneName);
+
+      do {
+        const requestParams = {
           StartTime: startTime,
           Filters: [{
             Name: 'product-description',
             Values: ['Linux/UNIX'],
           }],
-        }),
-      ]);
+        };
+        if (nextToken) {
+          requestParams.NextToken = nextToken;
+        }
 
-      // Get the information from the API responses that we care about
-      let zones = azResult.AvailabilityZones.map(x => x.ZoneName);
-      let pricePoints = priceResult.SpotPriceHistory.filter(x => zones.includes(x.AvailabilityZone));
+        let priceResult = await this.runaws(
+          ec2,
+          'describeSpotPriceHistory',
+          requestParams
+        );
 
-      // Determine what the prices which the API described.
-      let spotPrices = this._findSpotPricesForRegion({pricePoints});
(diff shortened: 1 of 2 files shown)


lib/pricing-poller.js
```diff
@@ -136,44 +136,49 @@ class PricingPoller {
...
+          Name: 'state',
+          Values: ['available'],
+        }],
+      });
+      let zones = azResult.AvailabilityZones.map(x => x.ZoneName);
+
+      do {
+        let priceResult = await this.runaws(ec2, 'describeSpotPriceHistory', {
           StartTime: startTime,
+          NextToken: nextToken,
```
jhford: I know it's probably paranoia, but in this code base I generally create an object and only the properties that are needed.  This is because a defined property with a value of undefined is different to a property that never has been defined, but returns undefined when dereferenced...  The JS object model is fun.

```js
> a = {}
{}
> a.b=undefined
undefined
> a
{ b: undefined }
>
> delete a.b
true
> a
{}
```

lib/pricing-poller.js
```diff
@@ -136,44 +136,49 @@ class PricingPoller {
...
-      // Get the information from the API responses that we care about
-      let zones = azResult.AvailabilityZones.map(x => x.ZoneName);
-      let pricePoints = priceResult.SpotPriceHistory.filter(x => zones.includes(x.AvailabilityZone));
+        nextToken = priceResult.NextToken;
 
-      // Determine what the prices which the API described.
-      let spotPrices = this._findSpotPricesForRegion({pricePoints});
-      let onDemandPrices = this._findOnDemandPricesForRegion();
+        // Get the information from the API responses that we care about
+        let pricePoints = priceResult.SpotPriceHistory.filter(x => zones.includes(x.AvailabilityZone));
```
jhford: heh, i couldn't figure out why you deleted `zones`, but it makes sense, since it was moved before the loop!

test/pricing-poller_test.js
```diff
@@ -130,7 +131,42 @@ describe('Pricing', () => {
       }],
     });
 
-    //describeSPHStub.onSecondCall().throws();
+    describeSPHStub.onSecondCall().callsFake((ec2, method, params) => {
+      assume(params.NextToken).is.equal('abc123');
+      return {
+        SpotPriceHistory: [{
+          Timestamp: '2017-07-05T13:11:21.000Z',
```
jhford: Can we bumb the timestamps and SpotPrices to ensure that there's a test case for a price that's from the first request and another from the second request?

Also, AvailabilityZone would probably be better in the full `us-east-1d` format, instead of `d`

lib/pricing-poller.js
```diff
@@ -136,44 +136,49 @@ class PricingPoller {
...
+
+        // Add some information to the on-demand prices
+        onDemandPrices = onDemandPrices.map(x => {
+          Object.assign(x, {type: 'ondemand', region});
+          return x;
+        });
 
-      Array.prototype.push.apply(prices, spotPrices);
-      Array.prototype.push.apply(prices, onDemandPrices);
+        Array.prototype.push.apply(prices, spotPrices);
```
jhford: Hmm, I think that since we're really looking for the max price of a given instanceType/region/zone triplet, we probably don't need to build one gigantic list in memory of all requests.  Instead, how about we find the maximum price of each api request, then compare it to the maximum price of the previous api request.

walac: Doesn't the call to `_findSpotPricesForRegion` do that?

PR #52, opened 2018-01-18
Update README to include PostgreSQL info

### LINKED ISSUE #51: README.md missing info about configuring database
The instructions in the readme to run tests (basically clone + `yarn` + `yarn test`) don't work because the tests are expecting a database to be in place.

It would be nice if the instructions in the readme actually worked.

indygreg:
This closes #51.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,49 @@ note is that the endpoint for submitting spot requests requires a fully formed
 and valid `LaunchSpecification`.
 
 ## Hacking
+
 ```
 git clone https://github.com/taskcluster/ec2-manager
 cd ec2-manager
-yarn
+yarn install --frozen-lockfile --ignore-optional
+yarn test
+```
+
+Tests require a working PostgreSQL server initialized with the
+schema of this project.
+
+To use Docker to start a PostgreSQL server:
+
+```
+docker run --rm -p 5432:5432 postgres:9.6
+```
+
+This will pull down and run a PostgreSQL server in the local terminal.
+It will map port 5432 in the container (the PostgreSQL server) to port
+5432 on the Docker host. The container will run in the foreground and take
+over the terminal. You can also use `docker run -d` to detach from the
+current terminal.
+
+Once the PostgreSQL server is running, create database schemas
+necessary to run the tests:
+
+```
+psql -h localhost -p 5432 -U postgres -f sql/create-db.sql
+```
+
+(The exact hostname - -h argument - may vary from machine to machine.)
+
+Tests will need the `DATABASE_URL` environment variable pointing to
+this PostgreSQL server. Set it to something like:
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,10 +61,49 @@ note is that the endpoint for submitting spot requests requires a fully formed
...
+yarn test
+```
+
+Tests require a working PostgreSQL server initialized with the
+schema of this project.
+
+To use Docker to start a PostgreSQL server:
+
+```
+docker run --rm -p 5432:5432 postgres:9.6
```
walac: I run with latest postgres, did you have any problem? Just curious

README.md
```diff
@@ -61,10 +61,49 @@ note is that the endpoint for submitting spot requests requires a fully formed
 and valid `LaunchSpecification`.
 
 ## Hacking
+
 ```
 git clone https://github.com/taskcluster/ec2-manager
 cd ec2-manager
-yarn
+yarn install --frozen-lockfile --ignore-optional
```
jhford: my understanding is that `--ignore-optional` is something related a specific OS X environment?  My understanding is that doing so would cause `libpg` and `pg-native` to be ignored.  Since that's what we're using in production, I'd really like to have it installed during development.  Maybe we could add a note to say to use it the flag if you have install troubles, but calling out the issue around missing `pg-native`?

README.md
```diff
@@ -61,10 +61,49 @@ note is that the endpoint for submitting spot requests requires a fully formed
...
+It will map port 5432 in the container (the PostgreSQL server) to port
+5432 on the Docker host. The container will run in the foreground and take
+over the terminal. You can also use `docker run -d` to detach from the
+current terminal.
+
+Once the PostgreSQL server is running, create database schemas
+necessary to run the tests:
+
+```
+psql -h localhost -p 5432 -U postgres -f sql/create-db.sql
```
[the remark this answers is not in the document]
jhford: Running the tests will also wipe out a development database, this command is useful to include because it shows how to recreate the DB for local dev usage.

README.md
```diff
@@ -61,10 +61,49 @@ note is that the endpoint for submitting spot requests requires a fully formed
...
+```
+
+(The exact hostname - -h argument - may vary from machine to machine.)
+
+Tests will need the `DATABASE_URL` environment variable pointing to
+this PostgreSQL server. Set it to something like:
+
+```
+export DATABASE_URL=postgres://postgres@localhost:5432/postgres
```
jhford: We could also note that someone could use `user-config.yml` to specify this instead of putting it into their environment.  That's what I do.

jhford (changes requested):
Looking great!  There's a couple things that I think we should change though.  Let me know if you agree.

imbstack:
Any chance this can be landed? I'm not sure what state it is in.

jhford:
Per the latest review feedback, we're close, but need some updates.

PR #54, opened 2018-01-25
Add the getCredentials and claimCredentials endpoint

walac:
`claimCrendentials` is used by the worker to get its initial credentials. The worker is validated using its [Instance Identity Document](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html), alongside its PKCS7 signature. 

The scopes given are the as those provided by [aws-provisioner](http://github.com/taskcluster/aws-provisioner).

@djmitche Could you please validate those scopes, please? Do we need to add or remove any scope?


lib/api.js
```diff
@@ -515,6 +521,62 @@ api.declare({
...
+  const sIdentDoc = new Buffer(req.body.identityDocument, 'base64').toString();
+  const signature = new Buffer(req.body.signature, 'base64').toString();
+
+  const E = () => {
+    return res.reportError('AuthorizationFailed', 'Failure to verify authorization');
+  };
+
+  if (!await validateIdentityDocument(sIdentDoc, signature)) {
+    return E();
+  }
```
djmitche: This is so cool :)

lib/api.js
```diff
@@ -515,6 +521,62 @@ api.declare({
...
+  if (instances.length != 1) {
+    return E();
+  }
+
+  const instance = instances[0];
+  if (instance.state != 'running' || instance.hasClaimedCredentials) {
+    return E();
+  }
+
+  const provisionerId = this.keyPrefix.slice(0, this.keyPrefix.length - 1);
```
djmitche: I think this patch is most of the work toward removing the need for a keyPrefix? Is it time to add a `provisionerId` config?

lib/api.js
```diff
@@ -515,6 +521,62 @@ api.declare({
...
+
+  const result = res.reply(taskcluster.createTemporaryCredentials({
+    scopes: [
+      `assume:worker-type:${provisionerId}/${instance.workerType}`,
+      'assume:worker-id:*',
+    ],
```
djmitche: Yes, these scopes look good.

lib/identity-document.js
```diff
@@ -0,0 +1,91 @@
...
+        err ? reject(err) : accept(fd);
+      });
+    });
+  }).then(fd => {
+    return new Promise((accept, reject) => {
+      fs.close(fd, err => {
+        err ? reject(err) : accept(tmp);
+      });
+    });
+  }).catch(err => tmp.unlik());
```
djmitche: `unlink`

Also, I think you want to re-reject the promise here?  Otherwise you'll just get back `undefined` from this promise.  `err => { tmp.unlink(); throw err; }`

Also, wouldn't https://nodejs.org/api/fs.html#fs_fs_writefilesync_file_data_options do the trick?

lib/api.js
```diff
@@ -515,6 +521,62 @@ api.declare({
...
+  input: 'claim-credentials-request.json#',
+  output: 'claim-credentials-response.json#',
+  description: [
+    'Yield credentials for the worker running in the given instance.',
+  ].join(' '),
+}, async function(req, res) {
+  const sIdentDoc = new Buffer(req.body.identityDocument, 'base64').toString();
+  const signature = new Buffer(req.body.signature, 'base64').toString();
+
+  const E = () => {
```
jhford: I think this name is a little confusing.  `reportError` would be much more clear.  I would like to log these errors with reasons in our logs, so that we can see why any particular request is failing.  Could we have this take an object and message for use in a `log.warn(obj, msg)` call?

lib/api.js
jhford: I think at this point, we should also strongly consider issuing
`assume:worker-id:$region:$instanceId` in place of the star scopes here.  Since this change is going to require a change to how workers work, this might be a good time to go through with that testing.  Both because we can, and because we finally have understanding that instance ids will not be reused.

lib/api.js
```diff
@@ -515,6 +521,62 @@ api.declare({
   res.reply(prices);
 });
 
+api.declare({
+  method: 'delete',
```
jhford: In order to operate well with retries, etc, we should have a separate GET and DELETE endpoint here.  The GET being used to obtain the credential and the DELETE is used to register receipt of the credentials.  This is the same flow we have for provisioner secrets as currently implemented.  Basically, the response handler running to completion without error is not a reliable signal for whether the worker type has received and understood the credential.  We don't want a machine that fails to get its credential be unable to retry and have to shut down.

lib/identity-document.js
```diff
@@ -0,0 +1,94 @@
+const {spawn} = require('child_process');
+const temporary = require('temporary');
+const fs = require('fs');
+
+const AWS_PUBLIC_CERTIFICATE =
```
jhford: I really like that you're building this into the code.  I think if this cert ever needed to change, the deployment overhead would justify the work.

lib/identity-document.js
```diff
@@ -0,0 +1,94 @@
...
+lRa2v1ntMX3caRVDdbtPEWmdxSCYsYFDk4mZrOLBA4GEAAKBgEbmeve5f8LIE/Gf
+MNmP9CM5eovQOGx5ho8WqD+aTebs+k2tn92BBPqeZqpWRa5P/+jrdKml1qx4llHW
+MXrs3IgIb6+hUIB+S8dz8/mmO0bpr76RoZVCXYab2CZedFut7qc3WUH9+EUAH5mw
+vSeDCOUMYQR7R9LINYwouHIziqQYMAkGByqGSM44BAMDLwAwLAIUWXBlk40xTwSw
+7HX32MxXYruse9ACFBNGmdX2ZBrVNGrN9N2f6ROk0k9K
+-----END CERTIFICATE-----`;
+
+// This function would be a lot simpler if we used bluebird
+// to promisify fs, we just don't want to add a new bloated
+// dependency only for such small piece of code.
```
jhford: Fwiw, we could use [`util.promisify()`](https://nodejs.org/api/util.html#util_util_promisify_original)

lib/identity-document.js
```diff
@@ -0,0 +1,94 @@
...
+//
+// fs module has the writeFile function, but tests showed that
+// it didn't flush data to disk before returning.
```
jhford: yikes!  Good catch.  that's expected because of how it's implemented, but it's so counter intuitive that a simplicity wrapper would include such a major problem.  I would expect that the wrapper do the safest thing by default and force use of fs.open and fs.write only to avoid doing fsync.

package.json
```diff
@@ -37,6 +37,7 @@
     "taskcluster-lib-monitor": "^4.6.3",
     "taskcluster-lib-testing": "^1.0.4",
     "taskcluster-lib-validate": "^2.1.0",
+    "temporary": "^0.0.8",
```
jhford: it's amazing that node doesn't have this built in.

lib/identity-document.js
```diff
@@ -0,0 +1,94 @@
...
+      });
+    });
+  }).then(fd => {
+    return new Promise((accept, reject) => {
+      fs.close(fd, err => {
+        err ? reject(err) : accept(tmp);
+      });
+    });
+  }).catch(err => {
+    tmp.unlink();
```
jhford: this is an async call which does not check the status of the operation.  We should either call the sync version or give it a callback.  The problem here is that if we stop deleting files we could bring down the service with a bunch of temporary files.

lib/api.js
jhford: agreed!

lib/identity-document.js
```diff
@@ -0,0 +1,94 @@
...
+    throw err;
+  });
+}
+
+// see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
+module.exports = async function(identityDoc, signature) {
```
jhford: I looked at this function and I agree that it definitely would accomplish the task.  I'm concerned that it might not be the optimal approach, however.  The concern is both overhead of spawning processes, but also that we're generating a bunch of files on disk and having to deal with the associated I/O complexity, which you've already encountered in tests.  It also makes us depend on having an openssl binary on the server, which could complicated future deployments.

I did some research and it looks like there's a broadly adopted crypto library called [forge](https://github.com/digitalbazaar/forge) that understands how to parse PKSC7 and the node's standard library [`crypto.createVerify`](https://nodejs.org/api/crypto.html#crypto_crypto_createverify_algorithm_options) function could be used to do the actual signature validation.  I would really like us to do this in-process using Node to do validation if we can.

lib/identity-document.js
```diff
@@ -0,0 +1,94 @@
...
+        '-inform',
+        'PEM',
+        '-content',
+        document.path,
+        '-certfile',
+        cert.path,
+        '-noverify',
+      ]);
+
+      openssl.on('close', exitCode => accept(!exitCode));
```
jhford: given the importance of this comparison to our security, I think we should consider rejecting the promise if the exit code is non-zero.

lib/api.js
djmitche: OK, and at that point we'd have a more straightforward way to get a provisionerId?  This just seems like an odd way to find that value.

lib/api.js
djmitche: That would be so awesome!!

lib/identity-document.js
jhford: I just chatted with @franziskuskiefer about this and his recommendation is that we not do crypto implemented in JS.  Since `crypto.createVerify` is implemented in the OpenSSL portion of the node library, we should be fine to use Forge for parsing the PKCS7 files and the node crypto library for the verification.  We should be careful to include notes that we not use the crypto portions of Forge, rather only the parsing portion.

jhford:
@walac how are things going?

walac:
@jhford I believe I fixed all the issues, could you please give another look at the patch?

sql/create-db.sql
```diff
@@ -37,6 +37,7 @@ CREATE TABLE IF NOT EXISTS instances (
                                     -- to ensure that we have correct ordering of cloud watch
                                     -- events
   touched TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  hasClaimedCredentials BOOLEAN NOT NULL DEFAULT FALSE,
```
jhford: please ensure that the `lib/state.js` methods for instances allow for creating instances with this value set explicitly to true or false.

test/api_test.js
```diff
@@ -569,6 +622,105 @@ describe('Api', () => {
...
+      } catch (err) {
+        return err.code;
+      }
+    };
+
+    const expectedError = 'AuthorizationFailed';
+    const createTemporaryCredentials = taskcluster.createTemporaryCredentials;
+
+    before(() => {
+      taskcluster.createTemporaryCredentials = sandbox.stub();
```
jhford: The reason why we're using the sinon sandbox is to avoid having to do these special cleanup-requiring things.  Please use the `sandbox.stub(.....)` call.  The overall `before()` and `after()` functions should already take care of the cleanup in the `after()` here.

test/api_test.js
```diff
@@ -569,6 +622,105 @@ describe('Api', () => {
...
+      const result = await client.getCredentials({
+        signature: pkcs7.signedData,
+      });
+      assume(result).eql({
+        clientId: 'myClientId',
+        accessToken: 'myAccessToken',
+        certificate: 'myCertificate',
+      });
+
+      /*await client.claimCredentials({
```
jhford: We should probably be calling this function for real in the test.  Is there any reason why it's commented out?  The whole commented out block is stuff that we must test.

lib/api.js
```diff
@@ -530,6 +536,93 @@ api.declare({
...
+  route: '/credentials',
+  name: 'claimCredentials',
+  title: 'Request worker credentials',
+  stability: API.stability.experimental,
+  input: 'claim-credentials-request.json#',
+  description: [
+    'Yield credentials for the worker running in the given instance.',
+  ].join(' '),
+}, async function(req, res) {
+  const reportError = (obj, msg) => {
```
jhford: Since there's only one call site for this function, we can probably remove it.

lib/api.js
```diff
@@ -530,6 +536,93 @@ api.declare({
...
+
+  const doc = getIdentityDocument(req.body.signature);
+
+  if (!doc) {
+    return reportError({
+      signature: req.body.signature,
+    }, 'Invalid signature');
+  }
+
+  const instances = await this.state.listInstances({
```
jhford: Let's move the state-based logic for answering the question of "is this instance in a state to generate credentials" into a `State.canClaimCredentials({id, region}, client)` method.

test/api_test.js
```diff
@@ -569,6 +622,105 @@ describe('Api', () => {
...
+        return err.code;
+      }
+    };
+
+    const expectedError = 'AuthorizationFailed';
+    const createTemporaryCredentials = taskcluster.createTemporaryCredentials;
+
+    before(() => {
+      taskcluster.createTemporaryCredentials = sandbox.stub();
+      taskcluster.createTemporaryCredentials.returns({
```
jhford: If we set a variable `expectedCredentials`, we could use that to set the return value of this stub as well as using it to ensure that the credentials to test against would be using the same object with `assume(actualCredentials).deeply.equals(expectedCredentials)`

test/api_test.js
```diff
@@ -569,6 +622,105 @@ describe('Api', () => {
...
+        clientId: 'myClientId',
+        accessToken: 'myAccessToken',
+        certificate: 'myCertificate',
+      });
+    });
+
+    after(() => {
+      taskcluster.createTemporaryCredentials = createTemporaryCredentials;
```
jhford: We shouldn't need to do this, rather we should be using the sandbox's `.restore()` functionality

test/api_test.js
walac: Good catch, I commented for a test and forgot to uncomment it

schemas/claim-credentials-request.yml
```diff
@@ -0,0 +1,14 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:                      "Get Credentials Request"
+description: |
+  An object with the Instance Identity Document and its corresponding signature
+type:                       object
+properties:
+  signature:
+    type: string
+    description:
+      The signature if the Identity Document, base64 encoded
```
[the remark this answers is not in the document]
walac: I don't know if I followed you on that, the `document` isn't needed because it is embedded in the signature. And if the signature is invalid, the information in the document isn't reliable. The base64 signature is what we get from Amazon.

lib/identity-document.js
```diff
@@ -0,0 +1,85 @@
...
+
+  const messageDigest = attrs.filter(attr => {
+    const attrType = asn1.derToOid(attr.value[0].value);
+    return attrType === messageDigestOid;
+  }).map(attr => attr.value[1].value[0].value)[0];
+
+  verify.update(buf);
+
+  const sig = new Buffer(p7.rawCapture.signature, 'binary');
+  if (h.digest('binary') !== messageDigest
```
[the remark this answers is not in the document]
jhford: I was talking to @franziskuskiefer about this and it might be better to not do a second hashing outside of the `verify` instance.

Also, looking at the metadata service, I saw that there's also an RSA2048 version of the signed message.  It's also stored in a PKCS#7 file and has a much stronger hashing algorithm (SHA256 vs SHA1).  We should use that.

lib/identity-document.js
```diff
@@ -0,0 +1,85 @@
...
+  const h = crypto.createHash(forge.oids[digestAlgorithmOid]);
+  const verify = crypto.createVerify(forge.oids[signatureAlgorithmOid]);
+  const messageDigestOid = forge.oids['messageDigest'];
+
+  h.update(doc);
+
+  const messageDigest = attrs.filter(attr => {
+    const attrType = asn1.derToOid(attr.value[0].value);
+    return attrType === messageDigestOid;
+  }).map(attr => attr.value[1].value[0].value)[0];
```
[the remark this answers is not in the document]
jhford: Does the forge library have named accessor methods for pkcs7 files?  It would be great if we didn't need to do specific indexing.

lib/identity-document.js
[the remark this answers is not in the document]
walac: afaic, if we don't do this hash, we don't validate anything. The `verify` part makes sure `messageDigest`  is valid and hashed by Amazon, while `h.digest` is what actually validates the message. If we skip this part, someone might tamper the Identity Document and we will never know.

lib/main.js
```diff
@@ -112,6 +112,7 @@ let load = loader({
           ec2: ec2,
           runaws: runaws,
           lsChecker: lsChecker,
+          credentials: cfg.taskcluster.credentials,
```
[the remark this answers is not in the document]
walac: Hrm, does it improve security somehow?

walac:
@jhford  I think I fixed most (all?) of the comments. Regarding SHA256, I looked at it but it has no official documentation, but I know [it is not PKCS7](https://forums.aws.amazon.com/thread.jspa?threadID=246161). Can we look at it in a follow-up PR?

One more thing: we currently depend on the node-forge master branch to make this work due to [this](https://github.com/digitalbazaar/forge/pull/566) and [this](https://github.com/digitalbazaar/forge/pull/565).

lib/identity-document.js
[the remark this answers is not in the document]
jhford: That's what the verify class does.  In order to verify the signature we need:

1. The document to verify
2. The algorithm to use for verification
3. The public key
4. The signature

These values are provided, respectively, by:

1. Calling `verify.update(buf)`
2. Passing in the correct algorithm to the `crypto.createVerify()` factory
3. Passing to the `verify.verify` function
4. Passing to the `verify.verify` function

The `messageDigest` is actually the hash of the message body, just a different name.  The thing that it's hashing is the document.  Also, it seems to me that the message content being stored in the PKCS7 file is some special thing that amazon does.  This is highlighted by their command to verify the signature requiring you to separately obtain the document and pass it along with the PCKS7 file to the openssl binary.  Let's not rely on this special behaviour.

Let's send a cleartext copy of the document and the PKCS7 file as well and do the verification, as tools expect.

I'm also having trouble getting the unit tests to pass, as well as on travis.  Are you able to run them locally?

jhford:
> @jhford I think I fixed most (all?) of the comments. Regarding SHA256, I looked at it but it has no official documentation, but I know it is not PKCS7. Can we look at it in a follow-up PR?

There's an `rsa2048` file in the instance metadata that is an RSA-SHA256 signature that we could use.  My understanding is that PKCS7 is a file format for storing information.  It's not inherently RSA or DSA, SHA1 or SHA256.  I think they just used the `/pkcs7` endpoint for whatever reason.

> One more thing: we currently depend on the node-forge master branch to make this work due to this and this.

Good to know.  Do you know when (if) they're going to ship those in a release?

lib/main.js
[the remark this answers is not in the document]
jhford: yes, by allowing us to have a different credential which is very locked down for issuing temporary credentials.  This would let us rotate that key without rotating the provisioner's key and vice versa.

lib/identity-document.js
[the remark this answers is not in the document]
walac: > The messageDigest is actually the hash of the message body, just a different name. The thing that it's hashing is the document.

Exactly. The signature verification only validates the `authenticatedAttributes` field, which contains the document hash. If we only verify that, we know that the digest in the signature is valid, but not the actual content.

> I'm also having trouble getting the unit tests to pass, as well as on Travis. Are you able to run them locally?

This is due at the time I pushed it, my forge patches weren't merged and I was relying upon an npm released version with patches applied locally. Not sure if we should wait for a release or freeze it master.

schemas/claim-credentials-request.yml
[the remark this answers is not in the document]
jhford: The document being embedded into the signature is an extension to the standard which seems to be unique to Amazon.  We should send just the pkcs7/rsa2048 file and the cleartext document.  The cleartext document should be base64 encoded to ensure that there's zero chance for JSON re-serialization errors, since any change to the file would cause verification to fail.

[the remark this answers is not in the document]
walac: Embedding the document in the signature is in the standard, just mark as optional. Amazon chose to embed it. But I see your point, will change that.

walac:
> There's a rsa2048 file in the instance metadata that is an RSA-SHA256 signature that we could use

Just to make sure we are on the same page, are you talking about http://169.254.169.254/latest/dynamic/instance-identity/signature ?

jhford:
@walac just a heads up that the library for doing verification is up for review at https://github.com/taskcluster/iid-verify/pull/1 when you have a chance :)

walac:
@jhford I believe I fixed all the problems, but Travis still fails and it feels like it is unrelated. Could you please confirm that?

jhford:
@walac I believe that the problem is there's something funky going on with the Postgres installation in travis.  I haven't ever been able to reproduce that failure locally or in heroku, but it happens on occasion in travis.  I will be adding extra logging to see if I can figure out what's going on.

test/state_test.js
```diff
@@ -518,6 +519,13 @@ describe('State', () => {
 
   });
 
+  it('should be able to claim credentials', async() => {
+    await db.insertInstance(defaultInst);
+    assume(await db.canClaimCredentials(defaultInst)).to.be.true;
```
jhford: This and `.false` need to be invoked with `()` or else they don't throw.
```
$ node
> let assume = require('assume')
undefined
> assume(false).to.be.true;
[Function: ok]
> assume(false).to.be.true();
Thrown: Error: Unknown assertation failure occured, assumed `false` to equal (===) true
```

jhford:
@walac I've taken another look and a lot of the concerns aren't addressed.  Maybe we should hop on vidyo and chat about it?  I can meet tomorrow at 16:30 CEST (Berlin) if you'd like.

walac:
@jhford hah, I see a lot of your comments don't show up for me in the conversation tab.

lib/api.js
```diff
@@ -543,6 +550,83 @@ api.declare({
...
+
+api.declare({
+  method: 'delete',
+  route: '/credentials',
+  name: 'claimCredentials',
+  title: 'Request worker credentials',
+  stability: API.stability.experimental,
+  input: 'claim-credentials-request.json#',
+  description: [
+    'Yield credentials for the worker running in the given instance.',
```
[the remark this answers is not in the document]
jhford: @walac and I had a quick chat about this over vidyo.  We decided that instead of doing any sort of scope protecting, we'd just resend the document and signature and use that to ensure the claimer is the right machine.  Since the trust is derived from these documents, they are as trustworthy as the taskcluster credentials.

This also makes worker implementation easier, it is a GET followed by a DELETE to the same API endpoint, which is much easier than a GET, parsing the document to build a DELETE URL and using TC credentials to make the request.

PR #56, opened 2018-02-05
Many improvements

jhford:
Here's three major commits to the ec2-manager.

### Polling DB for Terminations
This basically lets us know why instances shut down.  We're already able to track a couple things about instance terminations, mainly that they happen.  This patch adds a simple poller that will check every 100s (i think) for up to 200 terminations and update them in the database.

These are the error codes:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#api-error-codes-table-server

and they'll let us see why things which were running ended.  Since we're tracking instances terminate in good and bad states, as well as some metadata, we can do things like seeing whether a specific region/zone/type triplet is getting killed by amazon a lot.  We could also see things like the average time of a specific worker type.  Lots of oppurtunity for inspection!

Samples from my local testing:
```
testing=> select * from instances;
         id          | workerType |  region   |     az     | instanceType |     state     |   imageId    |        launched        |       lastEvent        |            touched            
---------------------+------------+-----------+------------+--------------+---------------+--------------+------------------------+------------------------+-------------------------------
 i-0e54d792b6eb7ab41 | tutorial   | us-west-1 | us-west-1c | m3.large     | shutting-down | ami-f85c1e98 | 2018-02-05 22:38:30+01 | 2018-02-05 22:40:19+01 | 2018-02-05 22:40:24.936512+01
(1 row)

testing=> select * from terminations;
 id | workerType | region | az | instanceType | imageId | code | reason | launched | terminated | lastEvent | touched 
----+------------+--------+----+--------------+---------+------+--------+----------+------------+-----------+---------
(0 rows)

testing=> select * from terminations;
         id          | workerType |  region   |     az     | instanceType |   imageId    | code | reason |        launched        |       terminated       |       lastEvent        |            touched            
---------------------+------------+-----------+------------+--------------+--------------+------+--------+------------------------+------------------------+------------------------+-------------------------------
 i-0e54d792b6eb7ab41 | tutorial   | us-west-1 | us-west-1c | m3.large     | ami-f85c1e98 |      |        | 2018-02-05 22:38:30+01 | 2018-02-05 22:40:53+01 | 2018-02-05 22:40:53+01 | 2018-02-05 22:40:54.715294+01
(1 row)

testing=> select * from terminations;
         id          | workerType |  region   |     az     | instanceType |   imageId    |             code             |         reason          |        launched        |       terminated       |         lastEvent          |            touched            
---------------------+------------+-----------+------------+--------------+--------------+------------------------------+-------------------------+------------------------+------------------------+----------------------------+-------------------------------
 i-0e54d792b6eb7ab41 | tutorial   | us-west-1 | us-west-1c | m3.large     | ami-f85c1e98 | Client.UserInitiatedShutdown | User initiated shutdown | 2018-02-05 22:38:30+01 | 2018-02-05 22:40:53+01 | 2018-02-05 22:41:25.379+01 | 2018-02-05 22:41:25.380666+01
(1 row)

testing=> 

```

### Remove key prefix concept from ec2-manager
Now that we've got atomic tag-on-create support for instances, we have no need for the ec2-manager to understand the janky key prefix system we used to use.  Yay!  '''NOTE''' that a corresponding provisioner patch will need to be written to change how SSH key creation is handled.  We might also wish to make SSH KeyPair Name a user-configurable value in the worker type definition, but with a default provisioner-wide setting.  Since this is >96h from when we deployed the tag-on-create, this should have no impact on running instances as running instances should have been created with accurate tags.

### Storing of AWS Request outcomes
This is mainly intended for runInstances, but it was the same amount of work for logging all requests.  There's *tons* of information that's super handy for figuring out the state of Amazon.  We can use this information to make better choices about which region/zone/type triplets are in good shape and which are to be avoided.  Since for special cased requests (basically runInstances), I'm tracking the workerType, we'll now be able to see things like the rate at which invalid parameters are submitted to the ec2 api.  We should expose this information.

Samples from my local testing:
```
testing=> select * from awsrequests where method = 'runInstances';
  region   |              requestId               |    duration     |    method    | service | error |           called           |         code          |                  message                   | workerType |     az     | instanceType |   imageId    
-----------+--------------------------------------+-----------------+--------------+---------+-------+----------------------------+-----------------------+--------------------------------------------+------------+------------+--------------+--------------
 us-west-1 | 0d051dd5-3931-4b70-8fa6-bb40ea8f4bbc | 00:00:01.433223 | runInstances | ec2     | t     | 2018-02-05 22:28:43.905+01 | InvalidParameterValue | Invalid value 'z3.large' for InstanceType. | tutorial   | us-west-1c | z3.large     | ami-f85c1e98
 us-west-1 | 61da2107-f7ac-4407-8800-f3eb26bb09c5 | 00:00:02.140156 | runInstances | ec2     | f     | 2018-02-05 22:30:55.146+01 |                       |                                            | tutorial   | us-west-1c | m3.large     | ami-f85c1e98
 us-west-1 | dbc51c67-93eb-446e-bbc3-8d744efaddb0 | 00:00:01.887676 | runInstances | ec2     | f     | 2018-02-05 22:38:29.005+01 |                       |                                            | tutorial   | us-west-1c | m3.large     | ami-f85c1e98
```

The aws-request stuff might look weird, because it is.  @jonasfj I submit this particular patch as an example of how annoying the AWS-SDK library is.  See the hack I need to be able to have a promise interface where there's a functional timeout and logging of request ids :)

Flagging @imbstack and @walac for review
CC @djmitche for thoughts on the SSH KeyPair configurability


lib/termination-poller.js
```diff
@@ -0,0 +1,140 @@
+const assert = require('assert');
+const {runAWSRequest} = require('./aws-request');
+const Iterate = require('taskcluster-lib-iterate');
+const log = require('./log');
+
+const INSTANCE_LIMIT = 200;
+const POLL_DELAY = 100; // I forget which units this is :/
```
djmitche: > `waitTime`: number of seconds between the conclusion of one iteration and commencement of another.

It's `waitTimeAfterFail` that's in ms.

lib/termination-poller.js
```diff
@@ -0,0 +1,140 @@
...
+    regions.forEach(region => assert(typeof region === 'string'));
+    assert(typeof state === 'object');
+
+    this.ec2 = ec2;
+    this.regions = regions;
+    this.state = state;
+    this.runaws = runaws;
+
+    this.iterator = new Iterate({
+      maxIterationTime: 1000 * 60 * 15,
```
djmitche: This is in seconds too, so remove the `1000 *`.

lib/termination-poller.js
```diff
@@ -0,0 +1,140 @@
...
+      watchDog: 1000 * 60 * 15,
```
djmitche: also in seconds

lib/termination-poller.js
```diff
@@ -0,0 +1,140 @@
...
+      maxFailures: 10,
+      waitTime: pollDelay,
+      handler: async(watchdog) => {
+        try {
+          await this.poll(watchdog);
+        } catch (err) {
+          console.dir(err.stack || err);
+          monitor.reportError(err, 'Error polling pricing data');
```
djmitche: Need to update the message :)

lib/termination-poller.js
```diff
@@ -0,0 +1,140 @@
...
+
+      // Let's find the instance ids for our region
+      let instanceIds = toPoll.filter(x => x.region === region).map(x => x.id);
+      log.info({instanceIds, region}, 'polling terminations');
+
+      // We don't want to bother with the rest of this function if there's no
+      // instance ids in this region
+      //if (instanceIds.length === 0) {
+        //return;
+      //}
```
djmitche: It's usually best to avoid checking in commented-out code.  Either remove it, or if it was unintentional, un-comment it.

lib/termination-poller.js
```diff
@@ -0,0 +1,140 @@
...
+      // fails we'll give up and try again
+      let result;
+      try {
+        result = await this.runaws(this.ec2[region], 'describeInstances', {
+          InstanceIds: instanceIds,
+        });
+      } catch (err) {
+        log.warn({err}, 'failed to describe instances in api');
+        throw err;
+        return;
```
djmitche: `throw` doesn't continue on to the next statement, so there's no need for a return here.

test/termination-poller_test.js
```diff
@@ -0,0 +1,134 @@
...
+      ec2,
+      state,
+      runaws: describeInstancesStub,
+      regions: cfg.app.regions,
+    });
+
+  });
+
+  after(async() => {
+    //await state._runScript('drop-db.sql');
```
djmitche: please uncomment this too :)

lib/api.js
```diff
@@ -282,117 +264,77 @@ api.declare({
...
-  return {
-    prefix: parts[0],
-    workerType: parts[1],
-    keyHash: parts[2],
-  };
-};
-
 api.declare({
   method: 'get',
-  route: '/worker-types/:workerType/key-pair',
```
djmitche: According to the logs, the provisioner calls this endpoint quite a bit, so changing the path will likely cause those calls to fail.

lib/api.js
```diff
@@ -282,117 +264,77 @@ api.declare({
...
   title: 'Ensure a KeyPair for a given worker type exists',
   stability: API.stability.experimental,
   deferAuth: true,
-  scopes: [['ec2-manager:manage-resources:<workerType>']],
+  input: 'create-key-pair.json#',
+  scopes: [['ec2-manager:manage-key-pairs:<name>']],
   description: [
     'Ensure that a keypair of a given name exists.  This call caches',
     'internally the list of keypair names it has ensured at least one',
     'time, and as such is safe to call repeatedly.  It is idempotent.',
```
djmitche: the code that caches known keypairs appears to have been removed, so this description should be adjusted accordingly.

lib/housekeeping.js
```diff
@@ -69,7 +69,39 @@ class HouseKeeper {
...
+            }
+
+            // If we don't own this instance, we just skip it outright
+            if (!managerId) {
+              log.error({instanceId: id}, 'missing Owner tag');
+              continue;
+            }
+
+            // If it doesn't have a manager id, we know it isn't owned by us
+            // and we skip it
```
djmitche: It looks like this comment and the one above are reversed?

lib/termination-poller.js
```diff
@@ -71,7 +76,7 @@ class TerminationPoller {
           InstanceIds: instanceIds,
         });
       } catch (err) {
-        log.warn({err}, 'failed to describe instances in api');
+        log.warn({err}, 'failed to describe instances for termination');
```
djmitche: It looks like this and a few other changes in this file should be cherry-picked from 93afe77 to 321c4c1

lib/api.js
jhford: yep, which is the provisioner patch that I was mentioning in the original message.

lib/termination-poller.js
jhford: the throw was debugging which shouldn've been deleted.

schemas/create-key-pair.yml
```diff
@@ -0,0 +1,13 @@
+$schema: http://json-schema.org/draft-04/schema#
+title: "ssh public key"
+type: object
+properties:
+  value:
```
imbstack: Maybe we call this `keypair` or `publicKey` or something?

lib/api.js
jhford: github is showing me comments out of order.

Wander, since the API is declared Experimental and the only known consumer is receiving a patch at the same time, there's no reason to do a staged deployment here.

lib/aws-request.js
```diff
@@ -24,75 +32,115 @@ const log = require('./log');
...
+  if (requestInfo.service === 'ec2') {
+    if (method === 'runInstances' && body) {
+      
+      if (body.TagSpecifications) {
+        loop1:
+        for (let spec of body.TagSpecifications || []) {
+          for (let tag of spec.Tags || []) {
+            if (tag.Key === 'Name') {
+              requestInfo.workerType = tag.Value;
+              break loop1;
```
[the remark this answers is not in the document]
jhford: nope, it's iterating over two collections, TagSpecifications and Tags.  The break is to shortcircuit once we've found the value we're looking for.  No reason to continue once we've found what we need.

lib/main.js
```diff
@@ -346,12 +346,25 @@ let load = loader({
     },
   },
 
+  terminationpoller: {
```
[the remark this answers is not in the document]
jhford: nope, all the module names here are all lower case.

lib/aws-request.js
```diff
@@ -24,75 +32,115 @@ const log = require('./log');
...
 
-    log.error(logObj, 'aws request failure');
-    throw err;
+    throw newErr;
+  } finally {
+    try {
+      await state.logAWSRequest(requestInfo);
+    } catch (err2) {
+      log.error({err: err2}, 'error reporting aws request');
+    }
```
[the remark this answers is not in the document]
jhford: i'd like to be consistent about using async/await instead of .then()/.catch().

lib/state.js
```diff
@@ -825,7 +825,111 @@ class State {
...
+      duration,
+      method,
+      service,
+      error,
+      called,
+    ];
+
+    if (code || message || error) {
+      assert(typeof code === 'string');
+      assert(typeof message === 'string');
```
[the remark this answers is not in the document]
jhford: it's saying that if you've given any of code, message or error, that you must have all and that code, message need to be strings and error must be boolean.  this is intentional.

lib/aws-request.js
```diff
@@ -24,75 +32,115 @@ const log = require('./log');
...
-    if (request.response && request.response.requestId) {
-      if (logObj.requestId) {
-        logObj.requestIdFromHack = request.response.requestId;
-      } else {
-        logObj.requestId = request.response.requestId;
+    if (!requestInfo.requestId) {
+      if (request && request.response && request.response.error) {
+        requestInfo.requestId = request.response.error.requestId;
+      } else if (request && request.response) {
+        requestInfo.requestId = request.response.requestId;
```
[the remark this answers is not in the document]
jhford: I think this is a case where a little more code makes the intent a lot more clear.

lib/state.js
```diff
@@ -825,7 +825,111 @@ class State {
...
+      assert(error);
```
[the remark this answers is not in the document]
jhford: nope, because you could have truthy `code`, but falsy `error`, which is what this is checking.  It's saying that if you have either a code or a message or it's an error, that you must have a code, a message and you must set error to true

lib/state.js
```diff
@@ -825,7 +825,111 @@ class State {
...
+      assert(typeof instanceType === 'string');
+      columns.push('instanceType');
+      values.push(instanceType);
+    }
+
+    if (imageId) {
+      assert(typeof imageId === 'string');
+      columns.push('imageId');
+      values.push(imageId);
+    }
```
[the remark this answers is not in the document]
jhford: yep, we could, but I think that makes intent less clear.

lib/state.js
```diff
@@ -825,7 +825,111 @@ class State {
...
+    let vals = [];
+    for (let i = 1 ; i <= values.length; i++) {
+      vals.push('$' + i);
+    }
+
+    // Fix the duration.  We want to store an interval, which requires a
+    // special values entry
+    let durIdx = columns.indexOf('duration');
+    values[durIdx] = duration;
+    vals[durIdx] = `$${durIdx + 1} * interval '1 us'`;
```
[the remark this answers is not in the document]
jhford: Thanks for the heads up.  That is oddly enough how you specify intervals in postgres.  It's not a traditional imperative language
```
testing=> select now(), now() + interval '1 day';
              now              |           ?column?            
-------------------------------+-------------------------------
 2018-02-07 14:55:17.167107+01 | 2018-02-08 14:55:17.167107+01
(1 row)

```

lib/termination-poller.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  stop() {
+    return this.iterator.stop();
+  }
+
+  async poll(watchdog, sharedState) {
+    let toPoll = await this.state.findTerminationsToPoll(INSTANCE_LIMIT);
+
+    if (toPoll.length && toPoll.length === 0) {
```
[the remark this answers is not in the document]
jhford: yep.  an even stronger check would be `typeof topoll.length === 'number'`, but since this function is one that we control, I'll just ensure that it's always going to return a list

lib/termination-poller.js
```diff
@@ -0,0 +1,141 @@
...
+    // Now let's update them.  We're not using any transactions here because we
+    // know that each instance's resolution should attempt update regardless of
+    // whether others fail.
+    for (let resolution of resolutions) {
+      try {
+        await this.state.updateTerminationState(resolution);
+      } catch (err) {
+        log.error({err, resolution}, 'error setting resolution reason');
+      }
+    }
```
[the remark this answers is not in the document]
jhford: theoretically, sure.  These queries take ~1ms each to operate and running possibly 200 at the same time could cause the client pool to become starved for more important tasks.

The whole termination poller does not have anything close to a realtime demand here, so doing them sequentially here isn't really a concern.

test/api_test.js
```diff
@@ -390,49 +390,38 @@ describe('Api', () => {
...
-      await client.ensureKeyPair(workerType);
+      }));
+      await client.ensureKeyPair('test', {value: 'ssh-rsa fakekey'});
       assume(runaws.callCount).equals(regions.length * 2);
       runaws.reset();
 
-      // Case 2
-      await client.ensureKeyPair(workerType);
-      assume(runaws.callCount).equals(0);
+      // Let's create a key pair, but find there's already one there
```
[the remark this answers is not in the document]
jhford: not a typo, but I'll rephrase.

lib/termination-poller.js
```diff
@@ -0,0 +1,140 @@
...
+    assert(Array.isArray(regions));
+    regions.forEach(region => assert(typeof region === 'string'));
+    assert(typeof state === 'object');
+
+    this.ec2 = ec2;
+    this.regions = regions;
+    this.state = state;
+    this.runaws = runaws;
+
+    this.iterator = new Iterate({
```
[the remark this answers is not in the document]
jhford: Yep!  There's lots of great stuff we could be putting into lib-monitor in this patch which I haven't done yet.  I was planning to do a general once-over to check where we could/should be using lib-monitor as lots has changed since that was originally implemented.

schemas/create-key-pair.yml
jhford: I was sort of annoyed that I couldn't just get a string as the complete request body, I went vague as a result :)

I'll rename i

jhford:
This was pushed to master branch commit 3f8ea66f0a2afb353df873ce2d83db5660d4b4f0

PR #58, opened 2018-02-13
Upgrade to scope expressions

imbstack:
Most of this is pretty straightforward but `terminateInstance` might require some close inspection to see if I have preserved the semantics properly!

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -58,15 +58,14 @@ api.declare({
   title: 'Run an instance',
   stability: API.stability.experimental,
   input: 'run-instance-request.json#',
-  scopes: [['ec2-manager:manage-resources:<workerType>']],
-  deferAuth: true,
+  scopes: 'ec2-manager:manage-resources:<workerType>',
   description: [
     'Request an instance of a worker type',
   ].join(' '),
 }, async function(req, res) {
   try {
     let workerType = req.params.workerType;
-    if (!req.satisfies({workerType: workerType})) { return undefined; }
+    await req.authorize({workerType: workerType});
     let {
       ClientToken,
       Region,
@@ -181,7 +180,7 @@ api.declare({
     assert(result.Instances.length === 1);
 
     let [instance] = result.Instances;
-    
+
     let opts = {
       workerType,
       region: Region,
@@ -221,16 +220,15 @@ api.declare({
   route: '/worker-types/:workerType/resources',
   name: 'terminateWorkerType',
   title: 'Terminate all resources from a worker type',
-  scopes: [['ec2-manager:manage-resources:<workerType>']],
+  scopes: 'ec2-manager:manage-resources:<workerType>',
   stability: API.stability.experimental,
-  deferAuth: true,
   description: [
     'Terminate all instances for this worker type',
   ].join(' '),
 }, async function(req, res) {
   let workerType = req.params.workerType;
 
-  if (!req.satisfies({workerType: workerType})) { return undefined; }
+  await req.authorize({workerType: workerType});
 
   let ids = await this.state.listIdsOfWorkerType({workerType});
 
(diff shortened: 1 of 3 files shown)


lib/api.js
```diff
@@ -57,15 +57,14 @@ api.declare({
...
-  deferAuth: true,
+  scopes: 'ec2-manager:manage-resources:<workerType>',
   description: [
     'Request an instance of a worker type',
   ].join(' '),
 }, async function(req, res) {
   try {
     let workerType = req.params.workerType;
-    if (!req.satisfies({workerType: workerType})) { return undefined; }
+    await req.authorize({workerType: workerType});
```
jhford: It's sort of scary to me that we're doing this as a promise, since forgetting an await on this promise would allow this code to silently not block the method from running, and would not generate an unhandled promise rejection error (I think).

lib/api.js
```diff
@@ -354,11 +350,13 @@ api.declare({
   name: 'terminateInstance',
   title: 'Terminate an instance',
   stability: API.stability.experimental,
-  deferAuth: true,
-  scopes: [
-    ['ec2-manager:manage-instances:<region>:<instanceId>'],
-    ['ec2-manager:manage-resources:<workerType>'],
-  ],
+  scopes: {AnyOf: [
```
jhford: i find this notation sort of confusing, so I'll have to defer to your judgement as to whether it's correct or not.

lib/api.js
```diff
@@ -354,11 +350,13 @@ api.declare({
...
+    'ec2-manager:manage-instances:<region>:<instanceId>',
+    {
+      if: 'hasWorkerType',
+      then: 'ec2-manager:manage-resources:<workerType>',
```
jhford: is this if limited to booleans or is it more expressive?

lib/api.js
```diff
@@ -376,15 +374,23 @@ api.declare({
...
   // should
-  if (!req.satisfies({region, instanceId})) { 
+  try {
+    await req.authorize({region, instanceId, hasWorkerType: false});
+  } catch (err) {
+    if (err.code !== 'AuthorizationError') {
+      throw err;
+    }
     let workerTypes = await this.state.listInstances({id: instanceId, region});
     switch (workerTypes.len != 1) {
```
jhford: not realted to this PR, but this should be a switch on `workerTypes.length`.  This is not a security issues because `workerTypes.len` should be an undefined property as the return value of that function is a list.  When `workerTypes.len` (undefined) is `!=`'d to `1`, it's true.  Because the switch is a switch on `true`, only the default case matches.

I can take a look at this later, but I'd like to fix this before we land this patch

See also:
```
> undefined != 1
true
> [].len
undefined
> switch (true) { case 0: console.log('zero'); break; case 1: console.log('one'); break; default: console.log('default')}
default
```

lib/api.js
imbstack: Yeah, it kinda scares me too. After a good bit of thinking on this,  I think it will actually create an `unhandledRejection` and we send alerts to sentry every time this happens so it will be caught quickly when it happens. For extra safety we could try adding https://github.com/mcollina/make-promises-safe as well?

lib/api.js
imbstack: Yeah, just limited to booleans just to keep the surface area of the expression language quite small.

lib/api.js
imbstack: I'm also trying to think of some other mechanism by which we can ensure that it is awaited but not having much success yet. Unfortunately it needs to talk to auth so it has to be async.

jhford:
This will be implemented as part of https://github.com/taskcluster/ec2-manager/pull/69
This PR served as a great basis for the scope expressions there, thanks.

PR #59, opened 2018-02-15
Improve how we call AWS endpoints

jhford:
Let's give up on the `.promise()` feature of the aws-sdk, it's more trouble than it's worth.  This patch makes it so that we're using the older style `.send()` and even listeners way of calling APIs.  There's also a change to timeout much quicker now, since almost all EC2 apis are called inside of tc-lib-api event handlers and those *must* be 30s or less.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -30,7 +30,7 @@ defaults:
     sslEnabled: true
     httpOptions:
       timeout: 10000
-      connect: 2000
+      connectTimeout: 2000
   postgres:
     databaseUrl: !env DATABASE_URL
   server:
@@ -60,7 +60,7 @@ production:
     env: production
     forceSSL: true
     trustProxy: true
-    baseUrl: 'https://ec2-manager.herokuapp.com/v1'
+    baseUrl: 'https://ec2-manager.taskcluster.net/v1'
   app:
     id: 'ec2-manager-production'
     publishMetadata: true
@@ -75,7 +75,7 @@ staging:
     env: production
     forceSSL: true
     trustProxy: true
-    baseUrl: 'https://ec2-manager-staging.herokuapp.com/v1'
+    baseUrl: 'https://ec2-manager-staging.taskcluster.net/v1'
   app:
     id: 'ec2-manager-staging'
     publishMetadata: false
--- a/config.yml
+++ b/config.yml
@@ -26,7 +26,11 @@ defaults:
   aws:
     accessKeyId: !env AWS_ACCESS_KEY_ID
     secretAccessKey: !env AWS_SECRET_ACCESS_KEY
-    maxRetries: 10
+    maxRetries: 3
+    sslEnabled: true
+    httpOptions:
+      timeout: 10000
+      connect: 2000
   postgres:
     databaseUrl: !env DATABASE_URL
   server:
--- a/lib/api.js
+++ b/lib/api.js
@@ -164,31 +164,10 @@ api.declare({
       log.warn({err}, 'Error while reporting AMI usage, ignoring');
     }
 
-    try {
-      await this.state.insertInstance(opts); 
+    await this.state.upsertInstance(opts); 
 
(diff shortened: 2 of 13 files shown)


imbstack:
Is there context on what the trouble with `.promise()` is? Should we stop using it in all of our services? It is quite commonly used.

jhford:
It does not work reliably and it's use makes debugging with AWS team difficult as it completely hides the ability to get requestIds for successful calls.

On reliability, the Promise.race() that was in this file was something that would cause a rejection after 2 minutes of not resolving or rejecting.  I have seen this get hit with decent regularity in production.  The AWS-SDK client is extremely complex, so I suspect something is dropping somewhere.

On debugging, `.promise()` makes it extremely difficult to get the requestId for anything that's not a failed request.  Those requestIds are really important when debugging certain issues with Amazon.  If you want to see the gymnastics required to get the requestId and still use .promise(), look at the `-`'d code :)

Basically, it looks like aws copied our aws-sdk-promised library and made it less useful when they included this functionality in the upstream library.

We should not use .promise() anywhere, in my experience.  I will look into making this a self-standing library that we can use elsewhere.

imbstack:
I've been thinking about this a bit and could we try to upstream fixes to the aws-sdk instead of fixing it in every service manually or creating a new tc-only lib?

jhford:
The aws-sdk client is extremely complex, and we've already set the options such that the timeouts should work as expected but they just don't with their promises.  I don't think diagnosing the aws-sdk client's promise support is a good use of our time, since the main issue here is broken timeouts, which are fairly easy to work around.  The older event-emitter based api is reliable.  This is a simple wrapper, less than 150 lines of code including comments, whitespace, imports and special logging features.

Pulling this into a library would essentially be establishing a calling convention for calls to AWS services.  The `.promise()` method is not reliable, and the alternative does not fit nicely into a promise-based sytem.  That's what this library would do: use the reliable calling convention (event emitter) to make reliable promises without having to re-invent the wheel in all of our services.

jhford:
@imbstack let me know if you have any other questions

imbstack:
Have we at least filed a bug on aws-sdk showing the issues we're having? Maybe there is a quick fix they can do?

I really don't want to stop using this really important library in the way that the library authors say it can be used.

jhford:
Well, we're still using it the way the author wants, we're just using it in a specific way, and providing tooling to simplify that usage.  The `.promise()` method that AWS added was basically a copy and paste of the taskcluster-specific `aws-sdk-promise` library to begin with.

imbstack:
Said in another way, I would feel much more comfortable with this if we put some effort in to find what the underlying issue is with aws-sdk since we use it so heavily.

jhford:
I've been working on a library which will solve this problem for us, in a general way that non-taskcluster users can make use of.  We should consider using that library to do our AWS calls elsewhere.  Maybe AWS will consider integrating the changes a second time.

If someone else has time for digging into it, I certainly won't hinder them.

jonasfj:
I agree with bstack that if we're deciding to not use a feature in aws-sdk because of some issue, we should report it.

That said, the `lib/aws-request.js` seems like a decent way to hide the horrors... I'm still not sure why it's smart to do transformations in there, but that seems to be a different question. Ie.`lib/aws-request.js` doesn't look like a generic wrapper around aws-sdk.

jhford:
I can file the bug, sure.

It doesn't present a generic wrapper because this is an ec2-manager
specific wrapper.  I'm working on a generic version that I intend to ship
and replace this with at some point.

jhford:
it looks like the push tests are passing, but the PR tests are failing. The build that's failing is the result of doing a merge by travis and testing *that*.  Since the actual push is reliably passing tests, I'll make sure that is what gets pushed to master branch (probably needs a rebase) and run those tests before I finish the push to master to be doubly sure it's working.

PR #60, opened 2018-02-15
State query cleanup

jhford:
I still need to test this in staging, but this change removes a bunch of complexity as well as fixing what could've been a small, edge-case bug in how state is tracked.

### DIFF
--- a/lib/cloud-watch-event-listener.js
+++ b/lib/cloud-watch-event-listener.js
@@ -92,12 +92,7 @@ class CloudWatchEventListener extends events.EventEmitter {
 
     let dbResponse;
 
-    try {
-      dbResponse = await this.state.listInstances({region, id}, transaction);
-    } catch (err) {
-      await this.state.rollbackTransaction(transaction);
-      throw err;
-    }
+    dbResponse = await this.state.listInstances({region, id}, transaction);
 
     if (dbResponse.length > 1) {
       // for this case to be hit, there'd have to be a database schema problem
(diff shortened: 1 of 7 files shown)


lib/main.js
```diff
@@ -161,6 +161,10 @@ let load = loader({
         }
       }
 
+      if (process.env.FORCE_JS_PG_CLIENT) {
```
imbstack: Why not make this a config.yml thing?

lib/cloud-watch-event-listener.js
```diff
@@ -147,39 +142,33 @@ class CloudWatchEventListener extends events.EventEmitter {
       } 
 
       if (dbResponse.lastEvent && dbResponse.lastEvent < generated) {
-        try {
```
imbstack: I'm confused why we can get rid of these try-catch now.

lib/main.js
jhford: it's only useful to get better errors when a SQL query fails to parse, only really useful when doing development.  I'd rather not bake that into the config.yml, since we really should be doing all testing, everwhere with PG-Native.  Doing this is explicitly an attempt to make this environment variable non-discoverable.

I'd rather delete this from the code altogether than put it in config.yml

lib/cloud-watch-event-listener.js
jhford: Because I missed it while doing my pass through.  Thanks for catching it.

jhford: Hmm, thinking about it more, we can probably get rid of the dbResponse.lastEvent checking here altogether and rely on the upsert's `ON CONFLICT... WHERE instances.terminated < $9` thingy.

jhford:
@imbstack hey, thanks for the catch on the cloud watch event listener.  I've drastically simplified the function based on your spotting that, thanks!

PR #61, opened 2018-02-19
spell publishMetaData the jonas way

djmitche:
It's spelled this way in `main.js`

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -115,7 +115,7 @@ let load = loader({
         },
         validator: validator,
         authBaseUrl: cfg.taskcluster.authBaseUrl,
-        publish: cfg.app.publishMetaData,
+        publish: cfg.app.publishMetadata,
         baseUrl: cfg.server.publicUrl + '/v1',
         referencePrefix: 'ec2-manager/v1/api.json',
         aws: cfg.aws,


jhford:
This is not a bug.  Metadata is a single word, per the [Websters](https://www.merriam-webster.com/dictionary/metadata) and [Oxford](https://en.oxforddictionaries.com/definition/metadata) dictionaries.  Per [wikipedia](https://en.wikipedia.org/wiki/Camel_case):

> Camel case (stylized as camelCase or CamelCase; also known as camel caps or more formally as medial capitals) is the practice of writing compound words or phrases such that each word or abbreviation in the middle of the phrase begins with a capital letter, with no intervening spaces or punctuation.

Metadata is not a compounding of words, so it follows that it not have a capital 'd'.  If we do anything, it should be to normalise on the correct usage in all of our services.

jhford:
@djmitche I've pushed a commit to https://github.com/taskcluster/ec2-manager/pull/59 (976d8cb1df81a9f334382718e99d49df5ae86eb7) which uses the correct spelling in `lib/main.js`.  Thanks for finding this, I wonder if it's why we're not getting references being published.

PR #64, opened 2018-03-08
Make taskcluster-client-web linting pass

helfi92:
The tabs are making the lines too long in `taskcluster-client-web`. I need to have this patch merged before I can start implementing the UI for the awesomeness that @jhford created.


helfi92:
@jhford  Whenever you get the chance :)

PR #65, opened 2018-03-09
Add some indices to improve performance

jhford:


jhford:
@petemoore if you think this is OK, feel free to merge it

PR #69, opened 2018-07-31
August updates

jhford:
This should be everything needed to support root url in ec2-manager.  I'm not totally certain about the lib-monitor stuff, so I'd appreciate a once-over on that part of the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,20 @@ yarn
 yarn test
 ```
 
+You'll need to have postgres running on your local system.  Once you have
+Postgres installed on your system, the following script can be run to get a
+working unit test environment
+
+```bash
+createuser testing
+createdb testing
+cat > user-config.yml << EOF
+test:
+  postgres:
+    databaseUrl: "postgres://testing:testing@localhost:5432/testing"
+EOF
+```
+
 ## Deployment notes
 When deploying, keep in mind the following:
 
--- a/lib/api.js
+++ b/lib/api.js
@@ -58,15 +58,13 @@ builder.declare({
   title: 'Run an instance',
   stability: APIBuilder.stability.experimental,
   input: 'run-instance-request.yml',
-  //scopes: [['ec2-manager:manage-resources:<workerType>']],
-  //deferAuth: true,
+  scopes: 'ec2-manager:manage-resources:<workerType>',
   description: [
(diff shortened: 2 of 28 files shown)


jhford:
Note that I'm using this PR to debug some travis issues

imbstack:
new `tasklcluster-lib-testing` should have the fix you need!

imbstack:
The lib-monitor stuff looks correct so far from what I've seen.

jhford:
Brian, I'm still sorting out the travis tests (their postgres setup is... complicated), but the tests pass locally.  Could you take another look when you have a chance?

jhford:
I have no clue why these tests are failing in travis... the exact same tests work perfectly locally.

lib/api.js
```diff
@@ -429,15 +420,23 @@ api.declare({
   // instance is not managed, and should not be authorized.  If there's more
   // than one, the database is not providing the relational guarantees it
   // should
-  if (!req.satisfies({region, instanceId})) { 
+  try {
+    await req.authorize({region, instanceId, hasWorkerType: false});
```
imbstack: I think we were planning to eventually only allow `authorize` once since it will eventually be a round-trip to auth service and back.  Can we do the entire check inside the switch below?

README.md
```diff
@@ -70,6 +70,20 @@ yarn
 yarn test
 ```
 
+You'll need to have postgres running on your local system.  Once you have
+Postgres installed on your system, the following script can be run to get a
+working unit test environment
+
+```bash
```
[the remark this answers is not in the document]
jhford: I'd like to take care of that while building out the postgres layer for the queue, but I completely agree.  A nice development environment for the new stuff is a top priority for me.

lib/api.js
jhford: The problem is that we'd be trading a round trip to Auth for a round trip to the Postgres database.  I'm not sure that it would be a problem, but it's not a totally free lookup either.

What's the timeline on moving to one-authorize()-per-handler?

jhford:
I've filed #70 for the authorize() calls, so that we don't forget about it, but not block a release of ec2-manager.

jhford:
a60d226..a42d707  master -> master

PR #71, opened 2018-09-13
Fixes

jhford:
A couple small fixes as well as Heroku based CI


jhford:
According to my email, Brian approved the change, but I think the repo rules caused it to be dismissed because of the inconsequential new commit (2db546f), which turned an approved to a "commented".

@imbstack I assume removing the travis build icon which will never change is not a problem for you, so I'm going to save the time of a review cycle... let me know if you'd rather I revert it.

PR #74, opened 2018-10-08
Add the claimCredentials endpoint

walac:
claimCredentials receives a RSA2048 signature of the identity document
and returns the credentials for the worker.

The document is validated according using the iid-verify [1] package.

Informations about the instance, like instance-id and region
are extracted from the identity document.

The given instance must be in a running state, otherwise the endpoint
will fail.

[1] https://www.npmjs.com/package/iid-verify

### DIFF
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -6,6 +6,63 @@ const {builder} = require('../lib/api');
 const sinon = require('sinon');
 const uuid = require('uuid');
 
+const pkcs7TestData = {
+  signedData: `
+MIAGCSqGSIb3DQEHAqCAMIACAQExCzAJBgUrDgMCGgUAMIAGCSqGSIb3DQEHAaCAJIAEggHVewog
+ICJkZXZwYXlQcm9kdWN0Q29kZXMiIDogbnVsbCwKICAibWFya2V0cGxhY2VQcm9kdWN0Q29kZXMi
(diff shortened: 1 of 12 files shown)


test/api_test.js
```diff
@@ -576,6 +633,84 @@ describe('Api', () => {
...
+    let expectedCredentials = {
+      clientId: 'myClientId',
+      accessToken: 'myAccessToken',
+      certificate: 'myCertificate',
+    };
+
+    let doc = JSON.parse(Buffer.from(pkcs7TestData.doc, 'base64').toString());
+
+    before(() => {
+      taskcluster.createTemporaryCredentials = sandbox.stub();
```
jhford: For the sinon sandboxing to work (`sandbox.restore()`) this should to be done slightly differently.  In other places, we create stubs() directly because we're not stubbing APIs, instead using the stub directly.  For the sandboxing to restore the taskcluster module correctly, sinon itself must do the assignment so that it can track it to revert it on `.restore()`.  This code might fix it, but I haven't tested it.

```javascript
   let stub = sandbox.stub(taskcluster, 'createTemporaryCredentials');
   stub.returns(expectedCredentials);
```

jhford (commented):
Nice work!  There's a couple small changes, but I think we're really close to being able to land + deploy.

PR #75, opened 2018-11-20
Bug 1508395 - upgrade tc-lib-validate (and allow multiple yarn versions)

djmitche:


jhford (commented):
Per the [best practises](https://github.com/taskcluster/taskcluster-docs/blame/eecabc6a4aaa456cedb4e107c72dc6c627487a8d/src/manual/design/devel/best-practices/microservices.md#L18)

> Prefer to use the latest stable node version, and corresponding yarn version. These should both be locked and not specified as a range. This means in particular no ^ in front of either version spec. Encode this version both in package.json and in any CI configurations such as .taskcluster.yml

Please remove the change to using a range for yarn versions.

djmitche:
Ah, I had forgotten yarn was in the best practices.  Hopefully renovate can help us at least use the same version of yarn across all services :)

PR #76, opened 2018-11-21
Fix `yarn test`

walac:
Details of the fixes are in the commit messages.

### DIFF
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -23,6 +23,8 @@ describe('Api', () => {
   let dbWorks = false;
 
   before(async () => {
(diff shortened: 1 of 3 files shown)


test/api_test.js
```diff
@@ -23,6 +23,8 @@ describe('Api', () => {
   let dbWorks = false;
 
   before(async () => {
+    process.env.TASKCLUSTER_ROOT_URL = 'http://localhost:5555/';
```
jhford: Is there a reason to save this in `process.env`?  Wouldn't a local variable also work?

walac: I could not detect where it comes from, but if I don't set the environment variable it doesn't work.

PR #77, opened 2019-01-23
Bug 1522097 - [ec2-manager] Show error messages on the error endpoints instead of placeholder ----- HIDDEN -----

jhford:
Background is that we want to display error messages to inform users of what the errors relating to a particular worker type are.


lib/state.js
```diff
@@ -1092,6 +1092,7 @@ class State {
       'SELECT region, az, "instanceType", "workerType", code, called AS time, message',
       'FROM awsrequests',
       'WHERE error = true AND method = \'runInstances\'',
+      'AND code <> \'UnauthorizedOperation\'',
```
imbstack: I'm guessing that this is here to avoid leaking information about our iam policies? It might be a bit confusing that some errors "arbitrarily" will not show up in the UI. Is there another way we can handle this. It might be fine to just show them as well.

jhford:
This will obfuscate the specific messages while showing all other errors.

PR #81, opened 2019-04-02
Bug 1541042 - Terminate instance shows Internal Server Error

helfi92:
From https://github.com/taskcluster/ec2-manager/blob/4554b53c545629d27ce67e20fe8569406aa5147c/lib/api.js#L399-L405

A user with `ec2-manager:manage-resources:<workerType>` is currently not able to terminate an instance. This is probably because the status code is not 'AuthorizationError` and so it was throwing early instead of running https://github.com/taskcluster/ec2-manager/blob/4554b53c545629d27ce67e20fe8569406aa5147c/lib/api.js#L434.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -423,7 +423,7 @@ builder.declare({
   try {
     await req.authorize({region, instanceId, hasWorkerType: false});
   } catch (err) {
-    if (err.code !== 'AuthorizationError') {
+    if (err.code === 'AuthorizationError') {
       throw err;
     }
     let workerTypes = await this.state.listInstances({id: instanceId, region});


lib/api.js
```diff
@@ -423,7 +423,7 @@ builder.declare({
   try {
     await req.authorize({region, instanceId, hasWorkerType: false});
   } catch (err) {
-    if (err.code !== 'AuthorizationError') {
```
imbstack, line 426: If I understand the logic of this correctly, I think what we actually want is:

```js
if (err.code !== 'InsufficientScopes') {
```

imbstack (approved):
Does my change make sense to you? That is the code `lib-api` returns when a user doesn't have enough scopes.

With this change it will try again with the second set of permissions but will still throw other unexpected errors as we like.

imbstack (changes requested):
Oops, meant to request changes

